PR #18, opened 2018-01-16
chore: upgrade nuxt to 1.1.1

clarkdo:


egoist:
I think we don't need to update it? https://github.com/nuxt-community/create-nuxt-app/issues/15#issuecomment-356621188

clarkdo:
@egoist aggree 👍

PR #20, opened 2018-01-18
Fix deprecation warnings since nuxt 1.0.0

Oliboy50:
BTW, could someone guide me to the documentation which explains what's the difference between `client` and `server` in the `nuxt.config.js` file?

Is `client` used with `nuxt generate` whereas `server` is used with `nuxt build`?


Atinux:
Actually, `isClient` & `isServer` are deprecated in favour of `procces.client` and `process.server` only for application files (not `nuxt.config.js`)

PR #77, opened 2018-08-08
Add prettier

thminhVN:
We add prettier plugin to help beautify our code


Atinux:
Thank you @thminhVN 

But this should be an option instead of forcing users to use prettier.

PR #92, opened 2018-09-22
fix: tailwind dependency

remarks:
Tailwind v0.6.6 allows `@apply` rules in `.vue` single file components so I thought upgrading it as the base version would be beneficial.

However, there are syntax errors from the linter when implementing postcss imports inside components, so adding `<style lang="postcss">` fixes that. In doing that, we'd need to add that test in `nuxt.config.js` with [these rules](https://github.com/nuxt/nuxt.js/issues/3231#issuecomment-381885334) as well.

### DIFF
--- a/template/_package.json
+++ b/template/_package.json
@@ -51,7 +51,7 @@
     "eslint-loader": "^2.0.0",
     "eslint-plugin-vue": "^4.0.0"<% } %><% if (ui === 'tailwind') { %>,
     "autoprefixer": "^8.6.4",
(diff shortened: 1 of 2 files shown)


Atinux:
Could you add it as well?

remarks:
Actually, Nuxt automatically parses as PostCSS so it runs as expected without it. So this is ready to go. I added some sample CSS in `index.vue` to serve as an example.

Atinux:
That it awesome, thank you @remarks

PR #98, opened 2018-09-25
Fix linting thrown from default template when enabling prettier option, container centering when enabling tailwind option

liam-potter:
This PR should resolve the errors that are thrown by the default template when users enable the prettier option during project creation.

### DIFF
--- a/template/_package.json
+++ b/template/_package.json
@@ -50,6 +50,7 @@
     "eslint": "^5.0.1",
     "eslint-loader": "^2.0.0",
     "eslint-plugin-vue": "^4.0.0"<% } %><% if (prettier === 'yes') { %>,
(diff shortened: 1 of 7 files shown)


aldarund:
@liam-potter Nice catch :) What about prettier errors with all other templates e.g. vuetify, buefy etc. Did u checked it?

liam-potter:
@aldarund The only other framework template with CSS changes required for prettier was Vuetify - I have now updated the required files too.

I have also fixed https://github.com/nuxt-community/create-nuxt-app/issues/97 in this PR.

liam-potter:
https://github.com/nuxt/nuxt.js/issues/3933 will also be resolved by this (related to https://github.com/nuxt-community/create-nuxt-app/issues/97).

PR #107, opened 2018-10-03
More information about the installation method.

ali-master:
Hey there,

Since everyone doesn't know how they can create an instance Nuxt.js application with the create-nuxt-app then I did update the README and installation and usage section.

Thank you all.


manniL:
We prefer using it through `npx`. An additional section would be okay I think.

Issue #103, opened 2018-09-28
Please add Sails integration


jorcons:
Is possible to Add Sails Backend integration? I'm not able to do it. Thanks.

PR #112, opened 2018-10-12
[WIP] add sails

tohutohu:
I try to add Sails backend integration. #103 

I managed to move it. But, this is still a WIP.

There are some uneasy points to merge in some respects.
- How simple is it to be?
Several files and directories that exist in this template can be deleted. However, it may cause users to get lost when adding features.
Comments can also be deleted, but there is a problem as well.

- I have not written a test
I will write it as soon as possible.

- CLI behavior
With the addition of Sails' choice, the choices now work as shown below.
movecursor


manniL:
Stale

PR #119, opened 2018-10-14
Updated with Vuetify original material colors in plugins/vuetify.js

Codefa:


Codefa:
> Where from this orignal material colors? https://material.io/design/color/the-color-system.html#color-theme-creation here i see different ones

@aldarund the colors come from here [](url)https://github.com/vuetifyjs/nuxt/blob/master/template/assets/style/app.styl

PR #123, opened 2018-10-18
feat: update vuetify  to 1.3.0 with automatic tree shaking

aldarund:


Codefa:
@aldarund please take a look into #119

PR #126, opened 2018-10-21
Remove bootstrap as this is installed by bootstrap-vue already.

uniibu:
Basically just removes bootstrap from the dependency section as this is already installed by `bootstrap-vue` by default. 

Unless this specific version of bootstrap is used somewhere else, then i think it is safe to remove on the list.


pi0:
Hey @uniibu. Actually this was intentional so that exact bootstrap version will be pinned in the project. Bootstrap may make css inconsistencies during updates while BV is almost components wrappers and JS API. 

Thanks for your contribution ❤️

PR #128, opened 2018-10-23
Updated Vuetify default.vue

Codefa:


aldarund:
What this PR suppose to do? it just will introduce eslint & prettier warning. The only change that make sense i see here is a dynamic year

PR #130, opened 2018-10-23
fix yarn command in README

Dam720:
`yarn create nuxt-app <my-project>`
to
`yarn create-nuxt-app <my-project>`


aldarund:
Nothing to be changed. yarn create nuxt-app is how it supposed to be working. See docs https://yarnpkg.com/lang/en/docs/cli/create/

PR #169, opened 2019-01-29
Missing dependency

aparajita:
`glob` is imported by `sao.js`, but is not included in the dependencies. It works under `npm` because `npm` flattens the tree, but that is an artifact of `npm`'s current implementation. When installing `create-nuxt-app` with `pnpm` it fails.

In any case, direct dependencies should always be listed!

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "cross-spawn": "^6.0.5",
+    "glob": "^7.1.3",
     "lodash": "^4.17.4",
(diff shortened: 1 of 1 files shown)


manniL:
Completely right! Thanks :relaxed:

aparajita:
Thank you for responding so quickly! 🙏

Aperrix:
Use ``pnpx create-nuxt-app <yourapp>`` instead of ``npx create-nuxt-app <yourapp>``. It's working for me 😃

PR #172, opened 2019-01-29
Move commas to after module name

thetre97:
When creating an app with the current config, the file ends up like this:

```javascript
modules: [
    , // Linter does not like this...
    'nuxt-buefy',
    '@nuxtjs/pwa'
  ],
```

So this update would change the file to look like this:

```javascript
modules: [
    'nuxt-buefy',
    '@nuxtjs/pwa', // Last trailing comma
  ],
```


clarkdo:
The last comma may need to be trimmed in someway since it's against eslint rule `Unexpected trailing comma.`.

clarkdo:
merged this for now, since the trailing will be fixed by `eslint --fix`.

PR #173, opened 2019-01-31
Instruct people to delete README.md from /static before production deploy

koresar:
I realised that `/static/README.md` was deployed to 3 different websites of ours. It's easy overlook the fact, becoming security concern. 

This simple change might help other people stay more secure.


manniL:
@koresar Thanks for the PR ☺️ 
Though I don't see a security reason in the Markdown file (why would it be one?), I agree that a hint wouldn't hurt

PR #175, opened 2019-02-01
fix: require nuxt-edge instead of nuxt

sibbngheid:
If app is generated with `--edge` flag require `nuxt-edge` instead of `nuxt`. Else console outputs `Error: Cannot find module 'nuxt'`


clarkdo:
Thanks @sibbngheid

PR #207, opened 2019-03-23
add TypeScript integration

takehiro-takahashi:
Hello everyone ! 

Added TypeScript integration option for Merge nuxt / nuxt.js # 5079. This will allow you to use TypeScript soon. 

With the addition of the TypeScript selection, the choices are as follows:
スクリーンショット 2019-03-23 16 19 17

In addition, the version of Nuxt.js has been updated to 2.5.1 to support TS.


kevinmarrec:
Thanks @takehiro-takahashi ! I don't know if we should maybe wait before releasing this feature as we're planning to change again how is handled `tsconfig.json`.

/cc @pi0 WDYT ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "create-nuxt-app",
-  "version": "2.6.0",
+  "version": "2.6.1",
```
clarkdo: The version doesn’t need to be bumped, we will bump it before release

takehiro-takahashi: @clarkdo 
Sorry, I fixed it now.
https://github.com/nuxt/create-nuxt-app/pull/207/commits/3839792f6c7f4e45e30e923d92ec8280c2841e77

clarkdo:
@takehiro-takahashi Thank you so much for the contribution 

@takehiro-takahashi @kevinmarrec @pi0 I think if ts is selected, typescript related code should be changed, like .vue with ts lang.

kevinmarrec:
@clarkdo Yes you're right

takehiro-takahashi:
@clarkdo @kevinmarrec 

I also thought that it might be necessary.
However, considering the freedom of users, I thought that it would be sufficient for the environment to be able to use TS.
If you need the TS version of the source code, I will prepare it immediately.

takehiro-takahashi:
@clarkdo @kevinmarrec 

Added TS base of Vuetify and nuxt default. iview is not supported because iview's TS type definition file itself did not work.
I am glad that you confirm at the timing.

test/snapshots/index.test.js.md
```diff
@@ -33,7 +33,7 @@ Generated by [AVA](https://ava.li).
     {
       dependencies: {
         'cross-env': '^5.2.0',
-        nuxt: '^2.4.0',
+        nuxt: '^2.5.1',
```
pi0: Is version bumped by this PR? And if so why not using `2.6.1`?

template/_package.json
```diff
@@ -50,7 +50,9 @@
     "nuxt-buefy": "^0.3.2"<% } else if (ui === 'iview') { %>,
     "iview": "3.1.5"<% } %><% if (axios === 'yes') { %>,
     "@nuxtjs/axios": "^5.3.6"<% } %><% if (pwa === 'yes') { %>,
-    "@nuxtjs/pwa": "^2.6.0"<% } %><% if (ui === 'tachyons') { %>,
+    "@nuxtjs/pwa": "^2.6.0"<% } %><% if (typescript === 'yes') { %>,
+    "@nuxt/typescript": "^2.5.1",
```
pi0: This would be better to be in `devDependencies` and instead, adding `ts-node` and `typescript` to dependencies.

template/_package.json
```diff
@@ -42,15 +42,17 @@
...
     "vuetify-loader": "^1.2.1"<% } else if (ui === 'bulma') { %>,
     "@nuxtjs/bulma": "^1.2.1"<% } else if (ui === 'element-ui') { %>,
     "element-ui": "^2.4.11"<% } else if (ui === 'ant-design-vue') { %>,
     "ant-design-vue": "^1.1.10"<% } else if (ui === 'buefy') { %>,
     "nuxt-buefy": "^0.3.2"<% } else if (ui === 'iview') { %>,
     "iview": "3.1.5"<% } %><% if (axios === 'yes') { %>,
     "@nuxtjs/axios": "^5.3.6"<% } %><% if (pwa === 'yes') { %>,
-    "@nuxtjs/pwa": "^2.6.0"<% } %><% if (ui === 'tachyons') { %>,
+    "@nuxtjs/pwa": "^2.6.0"<% } %><% if (typescript === 'yes') { %>,
+    "ts-node": "^8.0.3",
```
pi0: I think `typescript` is a peer dependency of it (for the compiler). /cc @kevinmarrec

kevinmarrec: That's true, and as `typescript` is shipped in `@nuxt/typescript` which is installed as devDep, it will bring the issue that `typescript` won't be installed in production.

kevinmarrec:
@pi0 What do you think of the status of this PR ? It seems kind of ready to me, unless we change our typescript behavior again (which has highly chance to happen, regarding a potential `nuxt.config` `.ts` extension check to enable TS support)

pi0:
I would suggest holding on this PR until latest final breaking change.

robsontenorio:
@takehiro-takahashi  @pi0 Is it time?

kevinmarrec:
Sorry for delay, we're waiting last breaking changes (outsourcing typescript support in dedicated module & repository) to get this PR finally ready & bring this typescript integration.

florianrusch:
@kevinmarrec @pi0 @takehiro-takahashi Are we still waiting? 🙂 Is there anything we can help with?

kevinmarrec:
@florianrusch Well lof things have changed since the last status of this PR
Furthermore I'm not sure if we should showcase `vue-property-decorator` (Class API)
We should showcase either Object API or Function API : see 

https://typescript.nuxtjs.org/cookbook/components.html (SPOIL Nuxt 2.9 :D)

kevinmarrec:
@takehiro-takahashi Thank you so much for your time, but I'm closing in favor of https://github.com/nuxt/create-nuxt-app/pull/328

PR #211, opened 2019-03-28
Axios options reference link corrected

Abhith:
Changed axios options reference link from https://github.com/nuxt-community/axios-module#options to https://axios.nuxtjs.org/options


manniL:
Thank you for your contribution ☺️

PR #212, opened 2019-03-29
feat: integration with feathers v3 and feathers-cli

subhadeepdas91:
On selection of "feathers" option during server selection a feather 3 project is created which is compatible with feathers-cli, further feathers-cli can be used to generated components like services, authentication etc.

Feathers config and test directories are contained within server directory


template/_package.json
```diff
@@ -53,7 +66,8 @@
     "@nuxtjs/pwa": "^2.6.0"<% } %><% if (ui === 'tachyons') { %>,
     "tachyons": "^4.11.1"<% } %>
   },
-  "devDependencies": {
+  "devDependencies": { <% if (server === 'feathers') { %>
+    "@feathersjs/cli": "^3.8.5",<% } %>
```
clarkdo: "nodemon" is better to be first

template/_package.json
```diff
@@ -3,8 +3,14 @@
...
-  "private": true,
-  "scripts": {<% if (server === 'adonis') { %>
+  "private": true, <% if (server === 'feathers') { %>
+  "directories": {
+    "lib": "server",
+    "config": "server/config",
+    "test": "server/test"
+  }, <% } %>
+  "scripts": {<% if (server === 'feathers') { %>
+    "feathers": "feathers",<% } %><% if (server === 'adonis') { %>
```
clarkdo: Should be `else if`

subhadeepdas91: placing `else if` causes non execution of `else` block immediately after `if (server === 'adonis')`
resulting to non generation of `dev, build, start, generate` tasks

should I place the `feathers` task as last task (after `test`)?

template/_package.json
subhadeepdas91: Fixed in commit 9f9aa8e394dfb948e6ea533e402a9ec32e1b7534

template/_package.json
clarkdo: Oh, I gotcha. Yes, putting `if (server === 'feathers')` after `dev, build, start, generate` makes more sense

subhadeepdas91: Fixed in commit 1910cb0da92dbcbb76618c19b61afb78bc0682d2

clarkdo:
@subhadeepdas91 Could you please update the snap, I think it's ready to merge now.

template/_package.json
```diff
@@ -3,7 +3,12 @@
   "version": "1.0.0",
   "description": "<%= description %>",
   "author": "<%= author %>",
-  "private": true,
+  "private": true, <% if (server === 'feathers') { %>
```
clarkdo: The space after `,` can be removed.

subhadeepdas91: Unnecessary space removed on commit fd0f9c4eb844e0b26483b6bc9da6f8dd3cc410ee

clarkdo:
As a result of team decision, server frameworks have been removed https://github.com/nuxt/create-nuxt-app/pull/501

PR #216, opened 2019-04-08
feat: updgrade @nuxtjs/pwa to v3

takehiro-takahashi:
pwa 2.6.0 had problems.
I made an update to solve the problem.


vinayakkulkarni:
shouldn't dependencies be stable versions of themselves and not `beta` ones?

pi0:
@clarkdo It is stable (as of testing period no issues reported) -- just have to create a proper release note and publish :)

Issue #217, opened 2019-04-09
Element UI breaks fresh install


mansouryaacoubi:
Hello,

while trying to create a fresh nuxt app, Element UI makes some problems.

### Versions
npm/npx: **6.7.0**
node: **11.12.0**

### Selected options:
custom server framework: **none**
features to install: **all**
custom ui framework: **element-ui**
custom test framework: **none**
rendering mode: **SPA**

![image](https://user-images.githubusercontent.com/6348858/55801106-9c666180-5ad5-11e9-884e-907b54ec23e2.png)

It just stuck here with installing packages..

And when I abort the package installation process and do an npm i myself, everything seems to work until I try to run `npm start` or `npm run dev`.

![image](https://user-images.githubusercontent.com/6348858/55801350-29111f80-5ad6-11e9-93ed-001642dcde42.png)

mansouryaacoubi:
I found a solution for when the installation got stuck... I don't know why this is but it seems creating the project using the verbose flag creates the app instantly.
`npx create-nuxt-app kitchen-dashboard --verbose`

but the problem with the errors do still exist.

mansouryaacoubi:
The problem seems to be line 27 in nuxt.config.js:

`css: ['element-ui/lib/theme-chalk/index.css'],`

![image](https://user-images.githubusercontent.com/6348858/55801704-f0be1100-5ad6-11e9-8981-d3f62e11c111.png)

mansouryaacoubi:
@manniL I don't mind :D

mansouryaacoubi:
Unfortunately I found out that this has to be another problem. Everything works just fine working on the same drive as the node installation. For example:

Node installation on drive C:\
Project deployment on drive C:\
=> everything works just fine

Node installation on drive C:\
Project deployment on drive D:\
=> webpack seems to have a problem with the path resolution

![image](https://user-images.githubusercontent.com/6348858/55940708-f3d51080-5c40-11e9-9912-238018831495.png)

I am not sure whether this is webpack problem or another problem.

mansouryaacoubi:
So finally I found out, what the real problem is. Neither `create-nuxt-app` nor `webpack` are faulty.
The problem was using the Windows 10 `mklink` functionality.

I was accessing my development path through `C:\Users\Mansour\dev` which was a windows link to `D:\development`.
When I changed my directory in the `cmd` to `D:\development`, it works without any problems.
[...]

PR #218, opened 2019-04-09
WIP: Fix element ui global css

mansouryaacoubi:
Fix for https://github.com/nuxt/create-nuxt-app/issues/217

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "create-nuxt-app",
-  "version": "2.6.0",
+  "version": "2.6.1",
   "description": "Create a Nuxt.js App in seconds.",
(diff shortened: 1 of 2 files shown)


mansouryaacoubi:
@manniL here you go 👍

mansouryaacoubi:
Unfortunately I found out that this has to be another problem. Everything works just fine working on the same drive as the node installation. For example:

Node installation on drive C:\
Project deployment on drive C:\
=> everything works just fine

Node installation on drive C:\
Project deployment on drive D:\
=> webpack seems to have a problem with the path resolution

image

I am not sure whether this is webpack problem or another problem.

PR #235, opened 2019-05-13
chore: improve config

pi0:
- Use `npm_package_description` and `npm_package_name` environment variables instead of directly importing `package.json. (Also compatible with yarn)
  - Less boilerplate
  - Less mem usage
  - Prevent problems of serverless deployment where `package.json` is not being added to lambda bundle
   - --> Something I suggest is that we can even eliminate that in the flavor of a clean default string. It is more clear how config is working. /cc @nuxt/core-team 

- Add newline for axios config


clarkdo:
conflict needs to be solved

pi0:
@clarkdo done

PR #237, opened 2019-05-14
feat: upgrade tailwindcss to v1

ChangJoo-Park:
https://twitter.com/adamwathan/status/1127315259449139202

tailwindcss has been released on version 1.0.0!

:)

PR #239, opened 2019-05-17
Updates tailwind from 0.7.0 to v1.0.1

DominikAngerer:
Updates tailwind to v1.0.1. Uses default configuration mentioned here: https://tailwindcss.com/docs/upgrading-to-v1/#2-update-your-config-file


DominikAngerer:
Oh wow - didn't see #237 - should we add the default configuration mentioned in the upgrade to there instead?

template/frameworks/tailwind/tailwind.config.js
```diff
@@ -0,0 +1,504 @@
+/*
```
manniL: I suggest to not load the whole default config here but only the minimal setup.

manniL:
I'd stick with #237 for now. Thanks anyway! 👍

PR #241, opened 2019-05-20
feat: add purge css for tailwind template

theomjones:
It makes sense to add purge css to the tailwind template to reduce file size, the Tailwind docs recommend this.
I've used https://github.com/Developmint/nuxt-purgecss

Thanks!


ChangJoo-Park:
tailwindcss and purgecss is makes perfect, but is It need to be in create-nuxt-app?

jpsc:
I think we can pass on this PR since we have a new [nuxt module](https://github.com/nuxt-community/nuxt-tailwindcss) that comes with this.

[Related PR](https://github.com/nuxt/create-nuxt-app/pull/254).

theomjones:
Ok sounds good, I'll close :)

PR #263, opened 2019-06-21
🔧 add support for vim swap files

acidjazz:


pimlie:
Care to explain the use-case for this?

swp files are temporary files which are created when you edit a file and removed once you close a file. If your project contains swp files on committing you are probably doing something wrong. Or are you using some GUI editor which uses vi internally which allows you to commit while editing files?

acidjazz:
@pimlie there are lots of cases but my personal one is i might want to check in stuff while my editor is open, if the `swp` files are not ignored they pop up as new files, it's quite common with vim users

pimlie:
Thanks, which editor do you use? I only use vim in the console thats why I was a bit surprised 😄 

Maybe we could take the opportunity to gather some more (editor specific) swap/backup files to add? Eg `~*`, `*.bak` and `.vscode` comes to mind.

acidjazz:
@pimlie macvim , i document my setup [here](https://github.com/acidjazz/dotfiles)

[this](https://github.com/github/gitignore) might have some help with that

manniL:
Could make sense, but I'm not sure if it'd be solved better by using a [global .gitignore](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore)

acidjazz:
@manniL it's the exact same concept as `.vscode` and .idea` in the editors section.

manniL:
For me it's fine then :relaxed:

PR #264, opened 2019-06-21
feat(cli): use cac for cli

evillt:
```console
$ yarn create nuxt-app --help
...
success Installed "create-nuxt-app@2.7.1" with binaries:
      - create-nuxt-app
> Generating Nuxt.js project in /Users/evila/--help
? Project name (--help) 
```

The `target path` is `/User/evila/--help` obviously not right 🤔

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,20 +1,31 @@
 #!/usr/bin/env node
 const path = require('path')
 const sao = require('sao')
+const cac = require('cac')
 const chalk = require('chalk')
 const { version } = require('./package.json')
 
 const generator = path.resolve(__dirname, './')
-// In a custom directory or current directory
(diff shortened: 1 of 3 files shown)


clarkdo:
@evillt Thank you for the pr, I'm curious that what's the reason for choosing cac, is there any benefit comparing with commander or minimalist ?

evillt:
CAC is very similar to commander.

The main reason is SAO already taken CAC as a dependency([you can see here](https://github.com/saojs/sao/blob/master/package.json#L23)), reuse a node-module to keep `create-nuxt-app` the same download size.

How do you think 🤔

clarkdo:
Thanks

evillt:
It would be honer 💃

PR #270, opened 2019-06-27
chore: update @nuxtjs/eslint-module

ricardogobbosouza:


manniL:
Not necessary IMO as `^0.X.Y` always includes all versions starting with `0.`

ricardogobbosouza:
@manniL You're right!

manniL:
I was wrong here :see_no_evil: 	
This doesn't apply to < v1 versions, so
`^0.1.Y` won't over everything starting with 0 but only everything starting with `0.1`.

ricardogobbosouza:
@manniL truth :grin:
I'm going to release `v1.0.0`, I think it's stable now

manniL:
:+1: 	
Feel free to update the deps here then :relaxed:

PR #272, opened 2019-06-28
feat(chore): included a dedicated option to show up envinfo

### LINKED ISSUE #240: Print out local environment information on passing in a dedicated option (--info)
It is good to have a dedicated option which prints out the local environment information.
## Proposed API
`create-nuxt-app --info`

```bash
Environment Info:

  System:
    OS: Linux 4.18 Ubuntu 18.10 (Cosmic Cuttlefish)
    CPU: (2) x64 Intel(R) Pentium(R) CPU G2020 @ 2.90GHz
  Binaries:
    Node: 8.11.4 - /usr/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 72.0.3626.81
    Firefox: 65.0
```

I would be happy to work on this :clap:

jamesgeorge007:
Fixes #240

## Proposed API

`create-nuxt-app --info`

```bash
Environment Info:

  System:
    OS: Linux 4.18 Ubuntu 18.10 (Cosmic Cuttlefish)
    CPU: (4) x64 Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz
  Binaries:
    Node: 8.11.4 - /usr/bin/node
    Yarn: 1.17.0-0 - /usr/local/bin/yarn
    npm: 6.9.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 74.0.3729.157
    Firefox: 67.0.2
  npmGlobalPackages:
    create-nuxt-app: 2.8.0
```


jamesgeorge007:
cc @clarkdo

cli.js
```diff
@@ -12,7 +13,23 @@ const cli = cac('create-nuxt-app')
 cli
   .command('[out-dir] [options]', 'Generate in a custom directory or current directory')
   .option('--edge', 'To install `nuxt-edge` instead of `nuxt`')
-  .action((outDir = '.') => {
+  .option('--info', 'Print out debugging information relating to the local environment')
+  .action(async (outDir = '.') => {
+    const hasInfoArg = process.argv.slice(2)[0] !== '--edge'
```
manniL:
```suggestion
    const hasInfoArg = process.argv.slice(2)[0] !== '--edge'
```

Could we make the check more stable? :thinking: 	
Something like "if `--info` is included as an argument at all"?

cli.js
```diff
@@ -12,7 +13,23 @@ const cli = cac('create-nuxt-app')
...
+    if (hasInfoArg) {
```
manniL: Maybe we could extract the if body as a function?

cli.js
jamesgeorge007: `cac` takes care of any other arguments passed in and warns appropriately except `--edge` and `--info`.

cli.js
jamesgeorge007: @manniL What you intend is to have the implementation within a dedicated method?

manniL: Improving code readability :relaxed:

cli.js
manniL: Maybe I'm getting this wrong but:

In case there will be a third flag (say `--foo`) added in the future, people would have to touch that line of code again as they'd have to check against `--edge` and `foo`. So I'd probably explicitly check for `=== '--info'` :thinking:

jamesgeorge007: Gotcha :+1:

jamesgeorge007:
cc @manniL

cli.js
```diff
@@ -3,26 +3,47 @@ const path = require('path')
...
-    sao({ generator, outDir, logLevel: 2 })
-      .run()
-      .catch((err) => {
-        console.trace(err)
-        process.exit(1)
-      })
+    const hasInfoArg = process.argv.slice(2)[0] === '--info'
+    if (hasInfoArg) {
+      showEnvInfo()
+    } else {
```
manniL: How about 
```
if (hasInfoArg) {
 return showEnvInfo()
}
```

You can avoid the else then :)

jamesgeorge007: :+1:

jamesgeorge007:
cc @manniL

jamesgeorge007:
cc @manniL

evillt:
Maybe you can use the `action` second param `options.info` to instead of `process.xxxxxxxxx`

But what this feature actually for?🤔

jamesgeorge007:
@evillt I had given it a try, and it doesn't seem to work as expected. 

> But what this feature actually for? :thinking:
The dedicated `--info` option is intended to simplify bug reports submissions. There could be a section within the PR/issue template where in which the users can paste the results as provided of `create-nuxt-app --info`.

cli.js
```diff
@@ -3,16 +3,36 @@ const path = require('path')
...
+  console.log(result)
+  process.exit(1)
+}
+
 cli
   .command('[out-dir] [options]', 'Generate in a custom directory or current directory')
   .option('--edge', 'To install `nuxt-edge` instead of `nuxt`')
+  .option('--info', 'Print out debugging information relating to the local environment')
   .action((outDir = '.') => {
+    const hasInfoArg = process.argv.slice(2)[0] === '--info'
```
jamesgeorge007: const [, , arg] = process.argv;
const hasInfoArg = arg === '--info';

jamesgeorge007: Just an alternative approach :thinking:

PR #273, opened 2019-07-01
fix: Duplicate key 'devModules'

sgraewe:
## This PR does the following things:

- Fix: Duplicate key 'devModules' error if using ESLINT + TailwindCSS


template/nuxt/nuxt.config.js
```diff
@@ -75,14 +75,19 @@ module.exports = {
     '@/plugins/antd-ui'
     <%_ } _%>
   ],
-  <%_ if (eslint) { _%>
   /*
-  ** Nuxt.js devModules
+  ** Nuxt.js dev-modules
   */
   devModules: [
```
manniL: Good catch! Maybe we could remove the whole `devModules` key in the config object if neither eslint nor tailwind had been selected?

template/nuxt/nuxt.config.js
```diff
@@ -75,14 +75,19 @@ module.exports = {
     '@/plugins/antd-ui'
     <%_ } _%>
   ],
-  <%_ if (eslint) { _%>
   /*
-  ** Nuxt.js devModules
+  ** Nuxt.js dev-modules
```
manniL: Good catch! Maybe we could remove the whole devModules key in the config object if neither eslint nor tailwind had been selected?

sgraewe: Like this?

manniL: :+1:

manniL:
LGTM. Now the only thing left are the tests :+1:

template/nuxt/nuxt.config.js
clarkdo: I think it could be kept, since if we add more devModules in future, the checking will be painful. And keeping an empty `devModules` can be considered as a hint that user may leverage it.

clarkdo: I removed the checking for now, we can revert my change if there is any other good reason

PR #277, opened 2019-07-05
feat: add jsconfig.json file for VS Code users

### LINKED ISSUE #294: Intellisense does not work on VS Code when using imports with Webpack aliases
I wrote an article to explain this issue and how to fix it 😉:
https://lmichelin.fr/vscode-intellisense-jsconfig-vue-js-nuxt-js-absolute-imports/

Cause: No `jsconfig.json` file is generated by `create-nuxt-app`.

lmichelin:
Fixes #294 

This pull request is intended to improve VS Code IntelliSense on a Nuxt.js project that uses absolute imports.

I wrote an article to explain this feature:
https://lmichelin.fr/vscode-intellisense-jsconfig-vue-js-nuxt-js-absolute-imports/

Feedback would be very appreciated! 😉


kevinmarrec:
@clarkdo I think `jsconfig.json` is suitable for providing intellisense in JavaScript projects meanwhile `tsconfig.json` is suitable for TypeScript projects.

So this PR is right BUT we'll need to only provide `tsconfig.json` if user has selected TypeScript option. Also, for TypeScript, I still don't know how are gonna handle TypeScript versions of all pages/components for every UI framework and other options.

saofile.js
```diff
@@ -115,6 +115,12 @@ module.exports = {
         { name: 'Single Page App', value: 'spa' }
       ],
       default: 'universal'
+    },
+    {
+      name: 'jsconfig',
```
clarkdo: Maybe we can add a `checkbox` as `DevTools` which only includes .jsConfig for now and we can add more goodies in future.

lmichelin: Good idea ! I updated my PR accordingly :slightly_smiling_face:

lmichelin: Good idea! I updated my PR accordingly :slightly_smiling_face:

clarkdo:
Thanks 🎉

PR #226, opened 2019-05-01
refactor: update ava template

vinayakkulkarni:
Currently when you pull the `ava` template, you have to face all sorts of errors out of the box, like this:
```
TypeError: Super expression must either be null or a function
```

## This PR does the following things:
- Fixes the ugly `../../components/logo.vue` imports in the tests with `@/components/logo.vue`
- Uses Babel 7
- This commit (https://github.com/nuxt/create-nuxt-app/commit/b3677f2d0617af818ebbab947e3db3a5ba29773a) mitigates this issue #180 but unfortunately it needs to be discussed better.  _More info:_
    - https://github.com/JeffreyWay/laravel-mix/issues/1748#issuecomment-422460613
    - https://github.com/vuejs/vue-cli/issues/2128#issuecomment-474740380 
- Compiling using `@babel/register`, read more about it [here.](https://github.com/avajs/ava/blob/master/docs/recipes/babel.md#compile-sources)
- Updated [`require-extension-hooks-babel`](https://github.com/jackmellis/require-extension-hooks-babel) & [`require-extension-hooks-vue`](https://github.com/jackmellis/require-extension-hooks-vue) (provided by the awesome @jackmellis)
---
_Note: After a lot of research, i've finally managed to setup nuxt with ava correctly with all the latest features: [Check out the repo here](https://github.com/vinayakkulkarni/nuxt-ava-e2e-unit-testing)_

PR #283, opened 2019-07-11
refactor: upgrade ava template

vinayakkulkarni:
The default [_`ava`_](https://ava.li) template is b0rked. #226  PR fixes that.

This PR is better than #226 because it has updated version of `Ava` (from 1.x to 2.x) which brings lots of breaking changes while ensuring you get the latest & greatest from the creators of Ava.


template/frameworks/ava/ava.config.js
```diff
@@ -1,5 +1,7 @@
 export default {
-  require: [
-    './test/helpers/setup.js'
-  ]
+  require: ['./test/_setup.js'],
+  sources: ['**/*.{js,vue}'],
+  files: ['test/specs/**/*'],
```
vinayakkulkarni: Note: This will only test the specfiles. for multiple uses of ava, check out [this repo](https://github.com/vinayakkulkarni/nuxt-ava-e2e-unit-testing)

template/_package.json
```diff
@@ -140,10 +140,12 @@
     "jest": "^24.1.0",
     "vue-jest": "^3.0.3",
     <%_ } else if (test === 'ava') { _%>
-    "ava": "^1.0.1",
+    "@babel/preset-env": "^7.5.4",
```
clarkdo: Any special reason for not using ava default config https://github.com/avajs/ava/blob/master/docs/recipes/babel.md#configuring-babel ?

template/_package.json
```diff
@@ -140,10 +140,12 @@
...
+    "ava": "^2.2.0",
+    "babel-plugin-module-resolver": "^3.2.0",
```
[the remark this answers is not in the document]
vinayakkulkarni: This babel plugin basically enables user to use the [alias](https://nuxtjs.org/guide/directory-structure#aliases) in the ava tests

I could remove it and send a new PR with this functionality.

Issue #286, opened 2019-07-16
Default express integration is outdated


TrueCarry:
Current version of nuxt works poorly with express as main entry point. It doesn't show  build progress, can't connect to dev websockets before full build of app. Maybe in future problems will be more severe.
How about using express as server middleware, or with nuxt express module? I can make PR if any of that is acceptable.

Atinux:
Hi @TrueCarry,

Indeed it has to be updated, using a module (adding a serverMiddleware) is the best approach at the moment before introducing Nuxt services (and avoiding sharing same process for Nuxt renderer and API).

Your PR is more than welcome :)

PR #292, opened 2019-07-20
Update express to use server middleware

TrueCarry:
https://github.com/nuxt/create-nuxt-app/issues/286


TrueCarry:
I've tried to fix test, but for some reasons if I run `yarn test` locally with .snap file removed it generates new .md one with CRLF line endings.

template/_package.json
```diff
@@ -21,6 +21,11 @@
     "dev": "nodemon --watch app --watch bootstrap --watch config --watch .env -x node server.js",
     "build": "cross-env NODE_ENV=production node ./ace nuxtbuild",
     "start": "cross-env NODE_ENV=production node server.js"
+    <%_ } else if (server === 'express') { _%>
```
clarkdo: I think this part can be same as none part

Atinux: agree

TrueCarry: What do you mean? The only way to correctly start nuxt in which it reports build progress in browser is AFAIK `--exec nuxt dev`. You can't start nuxt with server middleware by invoking `node server/index.js`.

TrueCarry: I've reread your comments. There's problems with using existing configs for express.

With server === 'none' it just runs nuxt dev, without watching server folder. In that case it'll not update server files if you change them, because currently nuxt doesn't support recompilation of serverMiddleware.
If you use last `else` with `"cross-env NODE_ENV=development nodemon server/index.js --watch server",` it'll not work with server middleware, since we need to start nuxt, node file.

So if you want to use default option for express, you need to start nuxt with nodemon, I'm not sure it's needed if you don't have server middleware.

clarkdo: `serverMiddleware` has already been watch since nuxt 2.7 which is from https://github.com/nuxt/nuxt.js/pull/5681

TrueCarry: I checked it while making this pull request and changing content of server/index.js makes nuxt recompile something, but doesn't actually change result of request to express middleware.

clarkdo: @Atinux

TrueCarry: Okay, so I've made couple more tests.
image
If you export just express instance hot reloading works.
If you export object with path and handler hot reloading doesn't work.
I'll update pull. Maybe we should create issue in nuxt repository.

TrueCarry: image
Turns out outside from exporting app, you also need to explicitly include that file.

Atinux:
Do you mind opening another PR for this please? Sorry for the outstanding delay 😬 

I do believe that to support express as servermiddleware, we should consider having just an `api/index.js` file and adding it as server middleware.

PR #299, opened 2019-07-26
chore: bump nuxt version

aldarund:


aldarund:
@clarkdo its so that new users dont ask why i used CNA and i have in package.json 2.0.0 instead of latest one

clarkdo:
I prefer sending them to `semver` site. 😅

aldarund:
thats a point, but there another questions that they ask so its just to minimize it.
Whatever, do as you feel better, its not important issue to discuss it much :)

manniL:
Agree with @clarkdo here. Better educate the user about NPM constraints and semver instead (maybe directly write a FAQ in the README?)

aldarund:
I`m afraid the user who ask such question dont read readme :)

manniL:
Could be :man_shrugging: but that should not be part of our scope.

pi0:
Some suggestion: We may show a message after setup you created a project using latest nuxt x.y.z

clarkdo:
I’ll add the message

PR #303, opened 2019-07-28
feat: Add e2e testing support using Ava

vinayakkulkarni:
## This PR enables users to perform E2E testing the Nuxt app using Ava test runner.

- [x] Based on [Alternative configuration files](https://github.com/avajs/ava/blob/master/docs/06-configuration.md#alternative-configuration-files) option provided by the creators of Ava.


template/_package.json
```diff
@@ -15,6 +15,11 @@
     <%_ if (test !== 'none') { _%>
     "test": "<%= test %>",
     <%_ } _%>
+    <%_ if (test === 'ava') { _%>
```
clarkdo: `test === 'ava'` can be before `test !== 'none'`, and make `test !== 'none'` be `else if`

template/_package.json
```diff
@@ -28,7 +28,11 @@
     <%_ if (eslint) { _%>
     "lint": "eslint --ext .js,.vue --ignore-path .gitignore .",
     <%_ } _%>
-    <%_ if (test !== 'none') { _%>
+    <%_ if (test === 'ava') { _%>
+    "test": "<%= test %>",
+    "test:unit": "cross-env TEST=unit ava --config unit.config.js",
+    "test:e2e": "cross-env TEST=e2e ava --config e2e-tests.config.js",
```
clarkdo: `e2e-tests.config.js` or `e2e.config.js` ?

BTW, how do you think moving test related config files to `test` dir ?

vinayakkulkarni: Yep. It could also be flexible enough to make use of all types of testing provided by Ava.

I'll update the PR and let's get some feedback from others too?

vinayakkulkarni:
> I can see some dir are test and others are tests, can me keep consistency?

the reason for that was the `/test` directory holds the actual e2e & unit tests whereas the `/tests` directory holds the testing setup.

it could be `/test/setup` where the `e2e.config.js` & `unit.config.js` could stay.. would be much cleaner.

clarkdo:
The config file can be just in test root dir or each own test dir ?

vinayakkulkarni:
> The config file can be just in test root dir or each own test dir ?

OK, according to the Ava docs

> The configuration file _must_ be in the same directory as the package.json file. 

Hence the `e2e.config.js` & `unit.config.js` were created in the root dir rather than inside the test directory.

clarkdo:
If so, can we only reserve ava.config.js and use process.env.TEST to apply different configurations?

vinayakkulkarni:
With dedicated config files. It'll be better to modify / update individual testing settings rather than have one huge `ava.config.js` with if-else checks.

Plus, that's how its recommended in Ava docs.

clarkdo:
> With dedicated config files. It'll be better to modify / update individual testing settings rather than have one huge `ava.config.js` with if-else checks.
> 
> Plus, that's how its recommended in Ava docs.

If so, what the point for setting TEST for each script ?

vinayakkulkarni:
`cross-env` is required for two reasons:

1. https://github.com/avajs/ava/issues/2112
2.
https://github.com/vinayakkulkarni/nuxt-ava-e2e-unit-testing/blob/master/test/_setup.js

Usually, the `ava.setup.js` differs for unit & e2e too.

clarkdo:
Sorry for the mis-clicking, I just wondered that `unit.config.js` and `e2e.config.js` may be a little confused since they don't indicate that they're ava related config files.

vinayakkulkarni:
Isn't that a good thing? They could be used to setup a whole testing framework within the nuxt app if anyone wants to use it in that bifurcated way?

Maybe using jest someone wants to do unit testing and using karma they want to do e2e testing and these two files could be modified for that purpose?

clarkdo:
This pr can be merged after resolving conflicts

clarkdo:
Thanks!

PR #304, opened 2019-07-28
feat: add alias (@/~) support in ava tests

vinayakkulkarni:
- [x] Mentioned about this feature in PR #283 

Refer: https://github.com/nuxt/create-nuxt-app/pull/283#discussion_r305201318

Basically allows users to import Vue SFCs in tests using aliases rather than the `../../` (relative?) imports.


template/frameworks/ava/babel.config.js
```diff
@@ -0,0 +1,18 @@
+module.exports = {
```
clarkdo: Adding `babel.config.js` may have conflicts if other frameworks also needs it.

Can we put the config in `babel` of `ava.config.js` ? https://github.com/avajs/ava/blob/master/docs/recipes/babel.md#customize-how-ava-compiles-your-test-files

vinayakkulkarni: - [x] Fixed in 4c26ca1

clarkdo:
Conflict needs to be resolved

vinayakkulkarni:
> Conflict needs to be resolved

Done :)

vinayakkulkarni:
Thoughts on merging this?

PR #361, opened 2019-10-11
Should only ignore root logs directory

aokabin:
We don't ignore logs directory If a project has logging api. (like `pages/app/logs/index.vue`)

So I suggest logs directory should ignore only root logs directory.

Thank you!


clarkdo:
Thanks 😸

PR #385, opened 2019-11-28
Add `rules` to config to allow "npm run dev" to work out of the box.

zeg-io:
The current implementation prevents `npm run dev` from building the client if stylelint was selected.  Adding this one property enables it to build properly out of the box.


manniL:
Thanks for the contribution! We already merged https://github.com/nuxt/create-nuxt-app/pull/387 to migitage that.

Thanks anyway :pray:

PR #394, opened 2019-12-03
chore: fancy background 🎉

sibbngheid:
Preview:
fancynuxtbg

Actual code is not jumping like in preview 😉


Atinux:
Thank you very much for the PR but actually it makes more code to the user since really few people use such background in their applications.

PR #415, opened 2020-01-21
fix(templates): Vuetify copyright year

Adamkillander96:
The copyright year is now 2019, should be 2020.

### DIFF
--- a/template/frameworks/vuetify/layouts/default.vue
+++ b/template/frameworks/vuetify/layouts/default.vue
@@ -83,7 +83,7 @@
       :fixed="fixed"
       app
     >
-      <span>&copy; 2019</span>
+      <span>&copy; 2020</span>
     </v-footer>
   </v-app>
 </template>
--- a/template/frameworks/vuetify/layouts/default.vue
+++ b/template/frameworks/vuetify/layouts/default.vue
@@ -83,7 +83,7 @@
       :fixed="fixed"
       app
     >
-      <span>&copy; 2020</span>
+      <span>&copy; {{ copyrightYear }}</span>
     </v-footer>
   </v-app>
 </template>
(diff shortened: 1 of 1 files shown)


template/frameworks/vuetify/layouts/default.vue
```diff
@@ -83,7 +83,7 @@
       :fixed="fixed"
       app
     >
-      <span>&copy; 2019</span>
+      <span>&copy; 2020</span>
```
vinayakkulkarni: Why not use: `new Date().getFullYear()`

template/frameworks/vuetify/layouts/default.vue
```diff
@@ -110,7 +110,8 @@ export default {
       miniVariant: false,
       right: true,
       rightDrawer: false,
-      title: 'Vuetify.js'
+      title: 'Vuetify.js',
+      copyrightYear: new Date().getFullYear()
```
vinayakkulkarni: You don't need another extra var, the interpolation `{{ new Date().getFullYear() }}` will work too

template/frameworks/vuetify/layouts/default.vue
```diff
@@ -83,7 +83,7 @@
       :fixed="fixed"
       app
     >
-      <span>&copy; 2019</span>
+      <span>&copy; {{ copyrightYear }}</span>
```
vinayakkulkarni: Why not just `{{ new Date().getFullYear() }}` instead of extra variable in `data()`

vinayakkulkarni:
Also, the title doesn't follow commitlint guide, refer this: [commitlint docs](https://github.com/conventional-changelog/commitlint/blob/master/README.md)

PR #424, opened 2020-02-07
feat: add Vuesax framework ui

luisDanielRoviraContreras:
Vuesax template added

http://vuesax.com/


luisDanielRoviraContreras:
We want to add the v4 that will soon be stable

the v3 will soon be deprecated and I prefer that v4 be added

clarkdo:
Thanks @luisDanielRoviraContreras

Issue #421, opened 2020-02-04
Testing newly generated app fails with "No tests found in test/ava.setup.js"


yelskiy:
Is there additional configuration needed to run initial tests?

After generating a new app
```
$ npx create-nuxt-app test-ava

create-nuxt-app v2.12.0
✨  Generating Nuxt.js project in test-ava
? Project name test-ava
? Project description My well-made Nuxt.js project
? Author name Alex
? Choose the package manager Npm
? Choose UI framework None
? Choose custom server framework None (Recommended)
? Choose Nuxt.js modules 
? Choose linting tools ESLint, Prettier
? Choose test framework AVA
? Choose rendering mode Single Page App
? Choose development tools jsconfig.json (Recommended for VS Code)
```
Running a test results in an error complaining of missing tests
```
$ npm run test

TAP version 13
# No tests found in test/ava.setup.js, make sure to import "ava" at the top of your test file
not ok 1 - No tests found in test/ava.setup.js, make sure to import "ava" at the top of your test file
Download the Vue Devtools extension for a better development experience:
https://github.com/vuejs/vue-devtools
# specs › Logo › is a Vue instance
ok 2 - specs › Logo › is a Vue instance
1..1
# tests 1
# pass 1
# fail 1

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! test-ava@1.0.0 test: `ava`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the test-ava@1.0.0 test script.
```

clarkdo:
Can you try with v2.14.0 which has updated to ava v3 ?

yelskiy:
Hi @clarkdo !  I just updated to v2.14.0, and tested again, but got pretty much the same result.

```sh
$ npx create-nuxt-app test2

create-nuxt-app v2.14.0
✨  Generating Nuxt.js project in test2
...

$ npm run test

TAP version 13
# No tests found in test/ava.setup.js, make sure to import "ava" at the top of your test file
not ok 1 - No tests found in test/ava.setup.js, make sure to import "ava" at the top of your test file
Download the Vue Devtools extension for a better development experience:
https://github.com/vuejs/vue-devtools
# specs › Logo › is a Vue instance
ok 2 - specs › Logo › is a Vue instance
Download the Vue Devtools extension for a better development experience:
https://github.com/vuejs/vue-devtools
# e2e › index › before hook
[...]

PR #426, opened 2020-02-10
fix(ava): no tests found in test/ava.setup.js

clarkdo:
Fix https://github.com/nuxt/create-nuxt-app/issues/421


clarkdo:
@vinayakkulkarni Can you give me a review ?

template/frameworks/ava/ava.config.cjs
```diff
@@ -1,5 +1,6 @@
 module.exports = () => {
   return {
+    files: ['./test/**/*.spec.js'],
```
vinayakkulkarni: This will mean that for e2e as well as unit tests, users would have to have `.spec.js` in their file name

clarkdo: Actually I think every real project will have fixed conveiton for tests naming, like `*.spec.js` or `*.test.js`, but your concern also makes sense.

I've moved `setup` to helpers which will be ingore by ava.

PR #428, opened 2020-02-13
chore: update vuetify logo

kyamashiro:
Vuetify logo is blurred, so I made it looks better.
* before
before

* after
after

ref: https://github.com/vuetifyjs/vuetify/blob/master/packages/kitchen/src/assets/logo.svg


kyamashiro:
Thank you for your review.
I renamed file name of vuetify-logo image.

PR #429, opened 2020-02-14
feat: added config.js

figurluk:
I've added support for "@/.." namespaces in WebStorm/PhpStorm.

These IDEs recognize file config.js then Auto-import and other features with "@/.." namespaces works.


template/_config.js
```diff
@@ -0,0 +1,5 @@
+System.config({
+  paths: {
+    '@/*': './*'
```
manniL: Could you add the other aliases, e.g. `~`, as well? ☺️

template/_config.js
```diff
@@ -0,0 +1,5 @@
+System.config({
```
manniL: Why a `config.js` instead of a `webpack.config.js`? 🤔 (See https://twitter.com/TheAlexLichter/status/1082945828270964736)

figurluk: I don't know why, but webpack.config.js didn`t work for me. So I prefer a solution which was successful for me.

pi0: I think it is better to elaborate more on this. I was also able to use webpack.config for IntelliJ idea and should consider part (probably a lot) of users are using different IDEs so should consider that to have maximum coverage :)

template/_config.js
```diff
@@ -0,0 +1,6 @@
...
```
pi0: If it is supporting a js export, maybe we can somehow read the list of available aliases from nuxt?

figurluk: Could you be more specific with "available aliases from nuxt". What do you mean?

template/_config.js
figurluk: I think that we can't cover all possible IDE's. This PR was primary for IntelliJ so in first, you could merge solution for that. If people with different IDE will face this problem, they have a chance to do the same thing -> Design solution, create PR and enjoy the result.

pi0:
As of webstorm beta-2, they have nuxt.js support also supporting nuxt.config:

https://blog.jetbrains.com/webstorm/2020/07/webstorm-2020-2-beta-2/

Thanks for PR @figurluk :heart:

PR #430, opened 2020-02-16
feat: new choice for PurgeCSS

carlos-avila:
An attempt to start addressing this issue. Added a conditional for Vuetify and PurgeCSS supports whitelistPatterns to include classes that are being missed by the path.


template/nuxt/nuxt.config.js
```diff
@@ -159,6 +162,25 @@ module.exports = {
...
+  ** https://github.com/Developmint/nuxt-purgecss
+  */
+  purgeCSS: {
+    mode: 'postcss',
+    paths: [
+      'components/**/*.vue',
+      'layouts/**/*.vue',
+      'pages/**/*.vue',
+      'plugins/**/*.js',
+      './node_modules/vuetify/src/**/*.ts'
```
clarkdo: Is  this line included only when ui is vuetify?

carlos-avila: I’ll try to do that and request a new pull.

carlos-avila: Addressed this in an [update](https://github.com/nuxt/create-nuxt-app/pull/430/commits/4a5edfc6366015ea357b6a09143ea7051f3f83c9)

Atinux:

Actually, after discussion with the team, it's better to not include such module at the beginning but instead having an easier way to add a module from nuxt directly (ie: `npx nuxt add purgecss`)

PR #431, opened 2020-02-16
feat: new feature choice for imagemin

carlos-avila:


Atinux:

Actually, after discussion with the team, it's better to not include such module at the beginning but instead having an easier way to add a module from nuxt directly (ie: `npx nuxt add imaginemin`)

PR #434, opened 2020-02-21
feat: new feature choice for Style Resources

carlos-avila:


template/_package.json
```diff
@@ -119,6 +119,9 @@
     <%_ if (ui === 'tachyons') { _%>
     "tachyons": "^4.11.1",
     <%_ } _%>
+    <%_ if (styleResources) { _%>
+    "@nuxtjs/style-resources": "^1.0.0",
```
clarkdo: Can we put the package as alphabetical order ?

carlos-avila: Sorted the packages in alpha order and moved the conditional for tachyons up as part of the rest of the of 'ui ===' group.

Atinux:

Actually, after discussion with the team, it's better to not include such module at the beginning but instead having an easier way to add a module from nuxt directly (ie: `npx nuxt add style-resources`)

When bootstraping a Nuxt app, few people actually now these module and can confuse them 😬 

We really do appreciate your help 💚

carlos-avila:
That sounds very interesting. I'd like to help with something like that. Let me know where I should start.

PR #454, opened 2020-04-02
fix: Updated tailwindcss version

ktopouzi:
### Fixes
- Pump up tailwindcss module version to get the latest Tailwind 

In a recent build I made with nuxt-create-app I realized that tailwindcss version i still on 1.0.0 and that prevents users accessing features available in tailwind 1.2 such as grids. I think this would be a good time to pump the tailwindcss module version. What do you say?


pi0:
Hi thanks for making this PR. Actually both "^1.0.0" and "^1.4.0" constraints will resolve to same result (latest version before 2.0.0)

CNA, intentially keeps version pinned to "^x.0.0" to avoid unnecessary updates for non major releases. 

Read more about semver and caret: https://nodesource.com/blog/semver-tilde-and-caret/

ktopouzi:
> Hi thanks for making this PR. Actually both "^1.0.0" and "^1.4.0" constraints will resolve to same result (latest version before 2.0.0)
> 
> CNA, intentially keeps version pinned to "^x.0.0" to avoid unnecessary updates for non major releases.
> 
> Read more about semver and caret: https://nodesource.com/blog/semver-tilde-and-caret/

Hey there! Didn't know about that. Thanks for the info. Actually I faced an issue with tailwind css grid features and I realized that nuxt-community/tailwindcss-module updated to tailwinds 1.2 on  v1.3.2 and that's the reason I made the PR. I believe we can safely close it though :)

manniL:
As @pi0 already explained, updating constraints is not necessary in this case ☺️

Thanks for the contribution though! 🙏🏻

adrianoresende:
I regretted using create-nuxt-app which seems to me to be outdated. Tailwind is old.
=/

manniL:
> I regretted using create-nuxt-app which seems to me to be outdated. Tailwind is old.
> =/

Neither `create-nuxt-app` nor the delivered tailwind version is outdated.
Please ensure that *you* are using the most recent `create-nuxt-app` version

PR #467, opened 2020-04-12
feat(template): support package.json and handler

clarkdo:
Add `package.json` and `package.js` for each package.

With this pr, we can support
- Auto deps pr from renovate, then we can keep updating packages to latest version
- Isolate each template package into its own dir
- Easier to add a new framework and better control by user `package.js`


packages/create-nuxt-app/lib/saofile.js
```diff
@@ -129,17 +132,27 @@ module.exports = {
...
-        delete data.scripts['']
-        delete data.dependencies['']
-        delete data.devDependencies['']
-        return data
+        return { ...data, ...pkg.load(generator) }
       }
     })
 
+    // For compiling pacakge.json
+    actions.push({
```
Atinux, line 145: Typo here

packages/cna-template/template/nuxt/package.js
```diff
@@ -0,0 +1,75 @@
...
+      delete pkg.devDependencies['eslint-plugin-nuxt']
+    }
+    if (!lintStaged) {
+      delete pkg.husky
+      delete pkg['lint-staged']
+      delete pkg.devDependencies.husky
+      delete pkg.devDependencies['lint-staged']
+    }
+    if (!stylelint) {
+      lintStaged && delete pkg['lint-staged']['*.{css,vue']
```
NickBolles, line 35: Is this missing the trailing `}`?

NickBolles: @clarkdo

clarkdo: Thanks for the pointing.

clarkdo: Will be fixed in https://github.com/nuxt/create-nuxt-app/pull/493/

PR #449, opened 2020-03-20
Feat: Generate typescript package types and tests

NickBolles:
Resubmitting as a pr to the base repo. 

Original PR: https://github.com/ChangJoo-Park/create-nuxt-app/pull/1
But it didn't make it into the initial typescript support here: https://github.com/nuxt/create-nuxt-app/pull/449

## List of changes 
1. Create nuxt.config.ts
2. Update nuxt.config.ts with changes from nuxt.config.js
3. Create a generate component template helper to switch between javascript and typescript

    - Another approach, if there are more typscript differences, could be a seperate file for js and typescript. but that seemed unnecessary right now

4. Update the following components to generate typescript components using the generateComponent template helper
    - buefy
    - FrameVuerk
    - Iview
    - Vuesax
    - Vuetify
5. Add typescript files to eslint and lint-staged configs
6. fix tests for typescript runtime
7. Add vue-shims.d.ts file

### Issues from #448 - @davelsan

1. [Fix] File XXX is not a module (Vetur error)
 - vue-shims.d.ts included above
2. [Fix] Remove linting of unused vars
  - Add a fix by prefixing them with a `_`, which is the standard way to declare an unused var
3. [Fix] Do not use the base ESLint indent rule
 - No fix for this in this PR. I'm going to leave this for another issue
4. [Feat] Add TypeScript linting to package.json scripts
- Included above
5. [Feat] Add TypeScript support to Jest test files
- added @types/jest as suggested by @davelsan
6. [Feat] Port Nuxt config to TypeScript
- Included above

PR #491, opened 2020-04-21
fix(lint): disable eslint nuxt.config.js extend line

zalog:
`config` and `ctx` are not used, I think it would be ok if we disable eslint on this line.


Atinux:
I think we can remove extends and add a line in the comments block to https://nuxtjs.org/api/configuration-build/

zalog:
Tnks for your time guys.
Ok @Atinux , I'll come back with a commit soon.

zalog:
Sorry @clarkdo I'm not sure how can I do that, can you be more specific please? :)

NickBolles:
this is already fixed as part of #449

PR #501, opened 2020-04-29
feat(templates): deprecate server templates

clarkdo:
According to the discussion for core team meeting, as Nuxt has published new programmatic API and nuxt/functions will be released soon, so members think we suggest users using [serverMiddleware](https://nuxtjs.org/api/configuration-servermiddleware/) as server-side first choice and remove all server templates in `create-nuxt-app`.

TODO:

- serverMiddleware template example
- post or article for Nuxt official suggestion and guide /cc @Atinux @pi0


clarkdo:
@nuxt/core-team , could you give this pr a review ?

Atinux:
Good for me 🙌

robbylucia:
I see the suggestions for replacing what was removed in this PR:

> Use nuxt-render https://nuxtjs.org/api/nuxt-render or serverMiddleware https://nuxtjs.org/api/configuration-servermiddleware

But neither of these are as easy as selecting "Express" during initialization-- or is it? Will there be something re-added to create-nuxt-app to bring back the one-click setup for Nuxt apps?

wizardpisces:
Where can we see the detailed discussion about why removing the server template is necessary? I could see the necessity of this change... From my perspective, it's maybe another feature which only  increasing the mental burden  and nothing more

pi0:
Hi and sorry for not communicating properly on this change. Please continue following up #563.

PR #529, opened 2020-05-18
feat: prepare for Nuxt 2.13

Atinux:
- Add `target` option
- Add `components: true` in `nuxt.config.js`
- Remove components import in `pages/index.vue`
- Add `nuxt export` and `nuxt serve` commands
- Remove `nuxt generate` command
- Use `<PascalCase>` for Vue components


packages/cna-template/template/nuxt/package.json
```diff
@@ -3,7 +3,8 @@
     "dev": "nuxt",
     "build": "nuxt build",
     "start": "nuxt start",
-    "generate": "nuxt generate",
+    "export": "nuxt export",
```
pi0, line 6: Should we conditionally add comments if target is static?

Atinux: I was thinking of it but afraid that renovate bot won't be working anymore, also it gives the possibility to easily switch targets by having all the comments.

clarkdo: It can be done in `package.js` of same dir.

packages/cna-template/template/nuxt/nuxt.config.js
```diff
@@ -68,6 +73,11 @@ export default {
     <%_ } _%>
   ],
   /*
+  ** Auto import components
+  ** See https://nuxtjs.org/api/configuration-components
+  */
+  components: true,
```
clarkdo, line 79: Is Nuxt.js officially suggesting using auto-detected components now ?

Can this pass eslint no-undef rule ?

acidjazz: I like to see it as an option, could also be promptable in the create-nuxt-app script

Atinux: Actually it's really easy to disable, just set it to `false` ;)

PR #534, opened 2020-05-23
fix: Log.vue is not a module when chose TypeScript

nori3tsu:
I added `<script lang="ts">` block to components/Logo.vue because `File '~~~ Logo.vue' is not a module.` occurs in pages/index.vue when TypeScript is selected in `create-nuxt-app` command.

スクリーンショット 2020-05-23 11 08 02

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,6 @@
 _*.js
 node_modules
+packages/cna-template/template/nuxt/components/Logo.vue
 packages/cna-template/template/nuxt/pages/index.vue
 packages/cna-template/template/frameworks/iview/pages/index.vue
 packages/cna-template/template/frameworks/jest/jest.config.js
--- a/packages/cna-template/template/nuxt/components/Logo.vue
+++ b/packages/cna-template/template/nuxt/components/Logo.vue
@@ -17,6 +17,15 @@
     </g>
   </svg>
 </template>
+
+<%_ if (typescript) { _%>
+<script lang="ts">
+import Vue from 'vue'
+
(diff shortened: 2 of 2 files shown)


JE-lee:
I can't agree that it is the solution。

nori3tsu:
Hi @JE-lee 
Thank you for the response.

What do you disagree with?

It works fine without this solution, but Vetur, which is the standard when developing Vue.js with VS Code, does not recognize it as a TypeScript module and an error occurs.

JE-lee:
> Hi @JE-lee
> Thank you for the response.
> 
> What do you disagree with?
> 
> It works fine without this solution, but Vetur, which is the standard when developing Vue.js with VS Code, does not recognize it as a TypeScript module and an error occurs.

At first I thought it was a vetur problem, and then I went inside vetur's issues to find out if anyone had had a similar problem.

but finally i figure out what's the problem.
becauce the Logo.vue doesn't has a export, so ts can resolve it as a module。

when I add the default export in the Logo.vue, the problem was resolved
```html
<script lang="ts">
export default {}
</script>
```

you can try

nori3tsu:
Yes, I think your solution is fine.
However, I think it's uniform for Logo.vue to use Vue.extend because `page/index.vue` uses the Optional API.

This problem is certainly due to Vetur, but it is also a problem encountered by unspecified users. You can save those users by adding this simple code.

pi0:
Hi! Latest template is using components feature to auto import. Also using latest VSCode + Vetur (with typescritp) i don't have this issue anymore.

PR #536, opened 2020-05-27
feat(test): add Cypress test framework as CLI option

amirrustam:


clarkdo:
@amirrustam I just pushed commit for moving cypess eslint config to a separate file under `spec`

amirrustam:
thank @clarkdo. I'm actually going to remove the examples, mostly because they don't adhere to the default eslint rules. They were originally included due to that being the default experience of Cypress.

I have a couple of improvements on-the-way to this PR to improve TypeScript experience and the base Cypress setup. After that, I'll get this PR out of draft mode.

SirMishaa:
An update soon to add Cypress? That would be appreciated.

vinayakkulkarni:
I have another simpler version of cypress + nuxt, would it be cool @amirrustam if I open a PR ?

amirrustam:
I appreciate the offer @vinayakkulkarni, but hold off on that because we just released Cypress v5, and accordingly got some new updates coming for this PR.

clarkdo:
Close as this has been hold for a long time.

PR #540, opened 2020-06-05
feat: add content module

benjamincanac:


Atinux:
Would be nice to also add a content/hello.md file (with a quick getting started inside)

Atinux:
I think you can directly add into the `cna-template` and add a filter here: https://github.com/nuxt/create-nuxt-app/blob/master/packages/create-nuxt-app/lib/saofile.js#L56

PR #544, opened 2020-06-08
feat(template): remove dotenv module

clarkdo:
As @Atinux has mentioned in discord, we should remove dotenv module from create-nuxt-app now.


web2033:
Can we know why?

Atinux:
@web2033 this is related to the fact that Nuxt v2.13 will directly support `.env` and `runtimeConfig`, see https://github.com/nuxt/nuxt.js/pull/7312

Actually `@nuxtjs/dotenv` actually bring a security vulnerability for people adding secret keys in `.env` that will be forwarded to the Webpack bundle and then exposed to the end users.

Atinux:
Why we removed the @nuxtjs/dotenv module

web2033:
@Atinux I think you meant this -> https://nuxtjs.org/blog/moving-from-nuxtjs-dotenv-to-runtime-config

PR #564, opened 2020-07-02
fix: remove generate command

debs-obrien:
remove generate command from ReadMe and add export command as well as serve command for static sites


Atinux:
Actually we are going to try to use `nuxt static` that will build & export.

PR #570, opened 2020-07-11
fix: do not proceed if the target directory exists and is non-empty

### LINKED ISSUE #556: CNA should fail if the target directory is not empty
Say, we have got a directory `my-app` in the current path. Firing in `create-nuxt-app my-app` goes ahead generating the project files within that directory.

jamesgeorge007:
closes #556


clarkdo:
Thanks @jamesgeorge007

PR #577, opened 2020-07-23
fix: make WebdriverIO tests pass after app creation

astagi:


clarkdo:
Thanks @astagi

astagi:
Yw @clarkdo

PR #581, opened 2020-08-01
feat: Add Dockerize option to nuxt project

hamidne:
This PR adds Dockerfile and docker-compose.yml as devOps option 🐳


packages/cna-template/template/_Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Dockerfile
+FROM node:14-alpine
```
pi0, line 2: - I would suggest going with main (debian) image as default since some native deps cannot be compiled on alpine
- Also would be nice if we split build into two stages example https://github.com/nuxt/nuxt.js/blob/dev/examples/docker-build/Dockerfile#L22

packages/cna-template/template/_docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.7'
+
+services:
+  app:
+    build: .
+    image: <%= name %>/<%= name %>
+    container_name: <%= name %>
+    restart: always
+    volumes:
+      - /app/node_modules
```
pi0, line 10: What's the purpose of permanent storage for `node_modules`?

packages/create-nuxt-app/lib/prompts.js
```diff
@@ -112,5 +112,15 @@ module.exports = [
...
     ],
     default: []
+  },
+  {
+    name: 'devOps',
+    message: 'DevOps tools:',
+    type: 'list',
+    choices: [
+      { name: 'None', value: 'none' },
+      { name: 'Dockerize project', value: 'docker' }
```
pi0, line 122: Can be simplify `Docker` or probably question asked as an option when setting target? Deployments are usually not known when initiating a project.

packages/cna-template/template/_docker-compose.yml
HamidNE: I added this to cache `node_modules` packages and speed up container execution.
Which I think is extra and the docker itself does that.

packages/cna-template/template/_Dockerfile
HamidNE: The alpine version is a very small version. I have been using this config for a long time and it can install packages well, the only problem is that it does not have git by default, which is when we use a package that instead of npm from use git, to solve this problem, I install git manually.
I have also seen some users have problems because the output size of the container with the image node itself is large.

packages/create-nuxt-app/lib/prompts.js
HamidNE: I changed its name to `Docker`.

packages/cna-template/template/_Dockerfile
HamidNE: And in this case, if you know something better, do it.

vinayakkulkarni: alpine > *

also, at geospoc we use nginx to serve as it provides nice brotli support out-of-the-box,

this is a huge change on how deployment of nuxt apps work..

PS. this is the Dockerfile we use at GeoSpoc

```Dockerfile
# Create the container from the alpine linux image
FROM alpine:latest as stage0

# Add nodejs
RUN apk add --update --no-cache nodejs nodejs-npm

# Set the directory we want to run the next commands for
WORKDIR /frontend

# Copy package.json & package-lock.json
COPY package.json package.json
COPY package-lock.json package-lock.json

# Install the dependencies, can be commented out if you're running the same node version
RUN npm ci

# Copy our source code into the container
COPY . .

# run webpack and the build the project
RUN NODE_ENV=production npm run build

# copy the built app to our served directory
FROM nginx:alpine

# Copy frontend data to dir that'll host the static files
COPY --from=stage0 /frontend/dist /var/www/html

# Copy the respective nginx configuration files
COPY docker/app/nginx/nginx.conf /etc/nginx/nginx.conf

# Copy the wildcard certificates from certs to /etc/ssl
COPY docker/app/nginx/certs/* /etc/ssl/

# build-base will add gcc
RUN apk add --update --no-cache git bash curl make build-base pcre-dev zlib-dev \
    && mkdir ~/brotli && command="nginx -v" && nginxv=$( ${command} 2>&1 ) && nginxlocal=$(echo $nginxv | grep -o '[0-9.]*$') \
    && cd ~/brotli && curl -L "https://nginx.org/download/nginx-$(echo $nginxlocal).tar.gz" -o nginx.tar.gz && tar zxvf nginx.tar.gz && rm nginx.tar.gz \
    && git clone https://github.com/google/ngx_brotli.git && cd ngx_brotli && git submodule update --init \
    && cd ~/brotli/nginx-$(echo $nginxlocal) && ./configure --with-compat --add-dynamic-module=../ngx_brotli && make modules \
    && cp -r ./objs/*.so /usr/lib/nginx/modules/ \
    && rm -rf ~/brotli/ \
    && sed -i '1iload_module modules/ngx_http_brotli_filter_module.so; load_module modules/ngx_http_brotli_static_module.so;' /etc/nginx/nginx.conf

# Expose ports.
EXPOSE 80
EXPOSE 443

# start nginx and keep the process from backgrounding and the container from quitting
CMD ["nginx"]
```

staged builds for life!

pi0 (changes requested):
PR pending because we should prefer multistage images.There might be many different approaches to build a docker image for nuxt apps. It would be probably better if we focus on documentation explaining  how to deploy projects via docker and some common practices IMO

packages/create-nuxt-app/lib/prompts.js
```diff
@@ -112,5 +112,15 @@ module.exports = [
       { name: 'Semantic Pull Requests', value: 'semantic-pull-requests' }
     ],
     default: []
+  },
+  {
+    name: 'devOps',
```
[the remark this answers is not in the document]
HamidNE: The selection of this feature is optional and also due to the code structure, it is not possible to enable this feature based on the condition.

HamidNE:
Of course, making a CI/CD for each project is time consuming and different for each project, and therefore requires a Docker image with a different structure, but in small projects having a code sample can be very helpful.
You should also spend more time on related documentation and introduce different Docker image structures so that people can create their own Docker image as needed.

packages/create-nuxt-app/lib/prompts.js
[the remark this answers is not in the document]
scscgit: There most certainly are use-cases for deploying full static sites as Dockerized, which except for using nginx itself include also scenarios like using a reverse proxy to expose private services inside a k8s cluster (and possibly whitelisting/blacklisting their endpoints) when the site consumes their API, or to go around CORS issues if the APIs cannot be accessed directly (though in some cases https://axios.nuxtjs.org/options/#proxy may be used for this, but it can't be used if we "already" use nginx anyway)

scscgit:
2 years is a pretty long time, is there any plan to introduce a first iteration of the Docker image variant yet? I'd like to eventually integrate new improvements or workarounds, but it's difficult to open PRs if nothing has been merged yet. Can we get any information about what is currently being worked on? It would be great if we had a documentation for various use-cases, or even better, if the Dockerfile was self-documented with available next steps after providing users with multiple generation options. For example, dev mode or build vs. generate may differ (and so does SSR), and we need workarounds like `envsubst` if we want to pass environment variables to nginx. Kubernetes tools could be also added in the following iterations, and so could tools like podman to become more inclusive.
* It's non-trivial to create a multi-platform baseline that new projects can easily reuse, and there are plenty of additional DX improvements available, for example adding common docker use-cases as npm scripts (including stop & rm to clean the containers, parametrizing env using cross-env-shell with workarounds (e.g. `cross-env-shell docker run --name project_name -v '\\\"$INIT_CWD:/usr/src/app\\\"'`), referring to build image using $npm_package_version, pipeline to push the image, etc.).

The pros and cons of specific implementations (e.g. base docker images and their versions) could be listed somewhere as a source of truth to make it easier to suggest changes once the dependent technologies further evolve, so that users aren't stuck with ad-hoc decisions in their projects if they prefer zero-configuration.

PR #594, opened 2020-08-27
feat: use `core-js@3` by default

pi0:
It is a long time since `core-js@3` released but we are still using `core-js@2` by default to avoid breaking changes. This is going to make more problems for nuxt@2 users as more packages updated.

Since [2.14.0](https://github.com/nuxt/nuxt.js/releases/tag/v2.14.0) we auto detect core-js version by resolving installed version from root dir ([src](https://github.com/nuxt/nuxt.js/pull/7785/files#diff-d8071aa99dfe30c3c1c6135dfd903334R123)) so this is a non-breaking change encouraging users to use 3.x for newly created projects.

PS: PR also contains sorting of dependencies by yarn


clarkdo (commented):
~~Explicit `corejs: 3` config in `build.babel` is needed. ~~

Oh, not realized it's auto detected now.

clarkdo (approved):

Atinux (approved):

Issue #587, opened 2020-08-12
Creating a project within an existing repository causes a fatal git error


LloydTao:
Creating a project within an existing repository renders git unable to track the project in its entirety.

**git add**
```
$ git add .
error: 'project/' does not have a commit checked out
fatal: adding files failed
```

**git status**
```
$ git status
On branch master

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        project/

nothing added to commit but untracked files present (use "git add" to track)
```

An option needs to be added for the creation of the `.git/` directory.

This is likely related  #366, which has been ignored since October 2019:

> _1. Not everyone uses git (shocking, I know)_
> _2. Sometimes you might be creating the project in an existing repository._

azrikahar:
This is resolved by #596 right?

LloydTao:
@azrikahar yeah! i was meant to write `Closes #587` 😁 

![image](https://user-images.githubusercontent.com/41923457/91993938-e5076600-ed2d-11ea-87f6-96a389e46038.png)

closing now.

PR #596, opened 2020-08-30
feat: prompt for git

LloydTao:
# Feat - Prompt for Git

Issue #587 - Creating a project within an existing repository causes a fatal git error

## Description

This pull request will prompt the user to choose a version control system.

image

There's two main reasons for this:

1. Using `npx create-nuxt-app` in an existing repository fails, as two `.git` directories exist. This is especially prevalent in our projects where we have a Django backend and Nuxt frontend in the same repo.
2. Several types of VCS exist (e.g. SVN), and users shouldn't be forced to delete unwanted directories in order to implement their preference. 

This feature has been designed to easily allow another VCS to be added to the list.

## Changes

The `prompt.js` asks for a VCS preference in the form of a list.

image

The `saofile.js` will now only run `gitInit()` if the respective choice is selected.

image


LloydTao:
Tests pass on my machine 😣 

image

LloydTao:
Re-linting directly before pushing did the trick. Life of a Windows dev 😁

LloydTao:
I'm now realising that these tests are auto-generated 😂

LloydTao:
All good to review now 😜

packages/create-nuxt-app/lib/prompts.js
```diff
@@ -112,5 +112,15 @@ module.exports = [
...
+  },
+  {
+    name: 'vcs',
+    message: 'Version control system:',
+    type: 'list',
+    choices: [
+      { name: 'None', value: 'none' },
+      { name: 'Git', value: 'git' }
+    ],
+    default: 'none'
```
pi0, line 124: Maybe it would be better to keep default as `git` like before? :)

LloydTao: @pi0 sure! my reasoning here was that the other fields use `default: 'none'`, so it was an attempt to match those design patterns, but i understand why Git is a good exception 😼

the prompt now uses `git` as default, as well as putting that option at the top of the list

- Use Git as default VCS: https://github.com/nuxt/create-nuxt-app/pull/596/commits/2924cc836eaa0ee59c61a5658797c349e54ccad2

pi0 (approved):
LGTM. Thanks :heart:

LloydTao:
thank you. this was my first open source contribution to a mainstream repo, so cheers for the good experience 😼

PR #597, opened 2020-08-31
fix: updated layout to be in par with vuetify 2.x

sam-eah:
v-row is a wrapper component for v-col. It utilizes flex properties to control the layout and flow of its inner columns. [...] This is the 2.x replacement for v-layout in 1.x.
v-col is a content holder that must be a direct child of v-row. This is the 2.x replacement for v-flex in 1.x.
https://vuetifyjs.com/fr-FR/components/grids/#usage


kevinmarrec (approved):
Sounds good :)

clarkdo (approved):

PR #599, opened 2020-09-03
fix(template): bump typescript packages versions

kevinmarrec:
These changes prevent new projects to install two TypeScript versions.

See : https://github.com/nuxt/typescript/issues/441


kevinmarrec:
Tests not passing but sounds to be a common Windows issue going on in all test runs

pi0 (approved):

PR #604, opened 2020-09-08
feat: add support for dependabot

vinayakkulkarni:
- [x] When merged, this will enable support for `dependabot` as one of the devTools options


clarkdo (commented):
As we're using [renovate](https://github.com/renovatebot/renovate) and also this is kind of outside scope of Nuxt.js, so I'm not sure if it's proper to be merged.

Let's hear feedback from @Atinux and @pi0

vinayakkulkarni:
> As we're using renovate and also this is kind of outside scope of Nuxt.js, so I'm not sure if it's proper to be merged.
>
> Let's hear feedback from @Atinux and @pi0

This is basically for a layman user who wants to have `dependabot` support out-of-the-box (without the need of adding the `.github/dependabot.yml` file manually afterwards). This PR doesn't affect the `create-nuxt-app` repo or the nuxt ecosystem in any ways (which uses renovate)

packages/create-nuxt-app/lib/prompts.js
```diff
@@ -109,10 +109,19 @@ module.exports = [
     type: 'checkbox',
     choices: [
       { name: 'jsconfig.json (Recommended for VS Code if you\'re not using typescript)', value: 'jsconfig.json' },
-      { name: 'Semantic Pull Requests', value: 'semantic-pull-requests' }
+      { name: 'Semantic Pull Requests', value: 'semantic-pull-requests' },
+      { name: 'Dependabot (For auto-updating dependencies)', value: 'dependabot' }
```
pi0, line 113: Dependabot only supports github and is going to github built-in soon. While still major share for private repo usage is for gitlab. (https://github.com/dependabot/dependabot-core/issues/399). In that case, i'm more in favor or having [renovate](https://github.com/renovatebot/renovate) option first but this PR should be valid as while as we mention it is for **github only**

pi0 (changes requested):
We should mention GH only or switch to renovate

packages/create-nuxt-app/lib/prompts.js
vinayakkulkarni: I most definitely can update the text to notify users that it'll work only on GitHub :)

vinayakkulkarni: - [x] Fixed in 5cc555c

pi0 (approved):
:+1: Looks awesome thanks :green_heart:

clarkdo (approved):
Great, can you please update README.md as well ?

vinayakkulkarni:
> Great, can you please update README.md as well ?

- [x] Done in a0da578

PR #606, opened 2020-09-09
refactor: reformat nuxt config 💄

pi0:
- Use one line comments 
- Use go.nuxtjs.dev shortlinks (this also allows us pointing users to updated link)
- Use `server: false` and target only if not defaults used
- One empty line between sections
- Improve descriptions
- Don't depend on `npm_package_`

**Rendered:** https://github.com/nuxt/create-nuxt-app/blob/d81a083d49e5fbec2c09e2c52d5590170a71c819/packages/create-nuxt-app/test/snapshots/index.test.js.md

**Current redirects:** https://github.com/nuxt/nuxt-redirects/blob/master/_redirects


clarkdo (approved):

packages/cna-template/template/nuxt/nuxt.config.js
```diff
@@ -1,21 +1,16 @@
...
-  ** See https://nuxtjs.org/api/configuration-mode
-  */
-  mode: '<%= mode %>',
-  /*
-  ** Nuxt target
-  ** See https://nuxtjs.org/api/configuration-target
-  */
+<%_ } _%>export default {
+  <%_ if (mode === 'spa') { _%>
+  // Disable server-side rendering (https://go.nuxtjs.dev/config-mode)
```
pi0, line 5: We need a new link

Atinux: Like https://go.nuxtjs.dev/config-ssr ?

packages/cna-template/template/nuxt/nuxt.config.js
```diff
@@ -1,21 +1,16 @@
...
+  server: false,
```
Atinux, line 6:
```suggestion
  ssr: false,
```

packages/cna-template/template/nuxt/nuxt.config.js
pi0: Yes! Debbie just created page i think (not merged)

PR #617, opened 2020-09-17
fix: Fix TypeScript build error.

282Haniwa:
To build with TypeScript, you need ts-loader.
That's why I added ts-loader to package.json.


clarkdo:
@282Haniwa `ts-loader` is a dependency of [`@nuxt/typescript-build`](https://github.com/nuxt/typescript/blob/c144f279a234cc6fcbc78b625af2f151c3518d58/packages/typescript-build/package.json#L20)

clarkdo (commented):
@282Haniwa `ts-loader` is a dependency of [`@nuxt/typescript-build`](https://github.com/nuxt/typescript/blob/c144f279a234cc6fcbc78b625af2f151c3518d58/packages/typescript-build/package.json#L20)

pi0:
Agree with @clarkdo. This should be fixed with last release 😊

282Haniwa:
Thank you for your support.

PR #619, opened 2020-09-17
fix(vuetify): updated layout to be on par with vuetify 2.x

tenzint:
v-row is a wrapper component for v-col. It utilizes flex properties to control the layout and flow of its inner columns. [...] This is the 2.x replacement for v-layout in 1.x.
v-col is a content holder that must be a direct child of v-row. This is the 2.x replacement for v-flex in 1.x.
https://vuetifyjs.com/en/components/grids/#usage

Related: #597


manniL:
LGTM. One leftover is to change the snapshots so the tests are matching.

tenzint:
hi. I tried but I'm unable to figure out how to start on fixing the snapshots. Can you shine a light on this? I don't have prior experience with PR.

I do know that snapshot is like an image frame of a video.

Should I be doing a git pull to fix this?

tenzint:
```
yes. You have to run `jest --updateSnapshot`
```
@manniL  Sorry for the 3 days gap (I had the discord browser closed). I appreciate your help. Since I didn't install jest, I saw that I can run it with `npx`.
I tried running `npx jest --updateSnapshot` on all of:
    1. the root of create-nuxt-app
    2. under `packages/create-nuxt-app/test`
    3. under `packages/create-nuxt-app/test/snapshots`
-------------
I am getting some error messages on all the above directories so I'm guessing I don't know which directory I should be on.
I thought option 3. was the right directory since `index.test.js.snap` was under there.

manniL:
Hey 👋🏻
Running `yarn test:snapshot` should be enough. You don't need other deps installed besides the ones from create-nuxt-app itself 😋

tenzint:
With your help I'm able to pass the `npm run test:snap` locally on my computer.

However, the new commit fails the snapshot check here in the remote.

----------------------------------------------------------

I checked that I'm on the `patch-1` branch with `git branch`. After running the test:snap successfully, I noticed that `index.test.js.snap` has updated so I did a `git add .` on the root and committed it. Then I did `git push origin patch-1` and I saw new checks happening on this page.

Sadly, the 2 test failures are still here.

pi0:
Hi @tenzint. Thanks for PR. I've rebased with master and updated snapshot :+1: 

Thanks @manniL for review

Nameless9:
Thanks @pi0 

and Thanks @manniL

tenzint:
Thanks @pi0 and @manniL

PR #624, opened 2020-10-03
chore: use husky to enforce lint and updated snapshots before commit

vinayakkulkarni:
The thought process behind this PR is that, as a contributor, there might be a chance where I might miss updating snapshots or in some cases, the code isn't linted to adhere to the standards

If this PR is merged, whenever user (contributor) tries to commit, the pre-commit hooks from husky will be triggered and the commit will be checked before it's actually committed to the branch

for example:
Screenshot 2020-10-03 at 10 38 51 AM


clarkdo (approved):
Nice

pi0:
Thanks for PR @vinayakkulkarni :+1: 

I've made a small modification to auto update/stage snapshot files as well :)

PR #630, opened 2020-10-07
feat: use default prompts values on non-interactive mode

patzick:
Hey!
We have a CLI process, which uses `create-nuxt-app` in non-interactive mode to prepare a new empty project. Unfortunately, providing `--answers` causes that all default values from prompts are omitted, so it's kinda "all or nothing", so adding any value to prompts is a breaking change for CI process.

In this PR I jumped into a case, when `--answers` flag is provided. Then I extract all default answers from `prompts` and overwrite them only by values provided in `answers` flag.
So users are able now to invoke non-interactive mode just by passing an empty object and project with default values will be created.

One thing I wasn't sure how to approach was interpolated default values. I added `cliDefault` property inside the prompt object, to provide a static default value as, at this point, I had no access to these values.

Hope you'll like that :) cheers!


clarkdo (commented):
Looks this will be solved in saojs v2 by using `mock: true`.
https://saojs.org/#testing-prompts

@egoist Is that possible to make `use default prompts value and answers object` a formal option rather than mock for testing purpose ?

clarkdo:
@patzick BTW, you can import `create-nuxt-app/lib/prompts` to get default prompts values and customise the answers object for now.

patzick:
Hey, @clarkdo thanks for jumping in!
Hopefully, this can be removed after saojs upgrade. Still, for now, it's a recurring problem which is breaking our CLI, as every time new version of `create-nuxt-app` introduces new param it breaks whole app initialization on our side.

Solution with importing `prompts` is unfortunately not a suitable workaround for our case. We're using `npx create-nuxt-app` to have always the newest version, so even if I'd add this to CLI dependency and prepare default answers importing `create-nuxt-app/lib/prompts` then after new release we would still break the process until we'd release new CLI version with updated `create-nuxt-app` dependency. So the same problem we have now.
This would be okay if I could do something like this
```js
const myAnswers // combination of provided answers and defaults from create-nuxt-app/lib/prompts
const createNuxtApp = require('create-nuxt-app')
await createNuxtApp(myAnswers)
```
this would ensure we're always using the same version, but it cannot be done this way.

If you have any other suggestions for solving that problem, I'd love to listen, but for now, only this PR is solving this problem and allowing to create new nuxt project via npx passing always the same answers as param.

clarkdo:
Hi @patzick , I think you can specify npx -p create-nuxt-app  and the script you pasted, then you’ll always use latest create-nuxt-app.

clarkdo:
I ll review this shortly and also ask other team members review, if team agrees with this fix, I think we can merge it for now as a quick fix and remove when saojs released better solution.

clarkdo:
@nuxt/framework Can you please review and take a look at above comments ?

Atinux:
I like the idea but instead I would create another flag named `--defaults` instead of `--answers`.

```bash
npx create-nuxt-app my-app --defaults
```

patzick:
hey @Atinux flag `--answers` already exist, it's not a new flag. Creating `--defaults` flag doesn't do much, as you not always want to have "only default" settings. Sometimes you just want to put 2-3 settings by `--answers`.
What the problem is really about is that with every new param in CLI this is actually a breaking change, as you need to provide **all the answers**, CLI is currently unable to take default values in this case and it's a simple fix for that.

PR #632, opened 2020-10-07
feat: add route page and content fetch

debs-obrien:
@clarkdo I am not sure how to run it locally to see if it works so please can you check this out:

This is all from feedback from when using this during live streams with developers new to Nuxt. Here is what has been added:

1. improved wording for features as people don't always know what content is - also added desc for axios
2. added a route to a simple page as when we create the app we only show links to external site and would be nice to show an internal link using NuxtLink as this is new for many users starting out with Nuxt
3. If content is selected also add a content page which fetches the the content and renders it on the page. Really useful as if not the content folder is created and nobody knows what to do with it so having a working example will really help

Any suggestions and improvements welcome


clarkdo:
Nice, I’ll have a look later today.

clarkdo (approved):

PR #634, opened 2020-10-08
fix(feat): add cloudinary module as a module option

mayashavin:


Atinux:
Thank you @mayashavin 

Actually we decided to stop adding modules to the list for create-nuxt-app but instead work on a way to add Nuxt modules on the fly (`nuxt add @nuxtjs/cloudinary`).

mayashavin:
No problem. I'm closing this then :)

Issue #553, opened 2020-06-20
[Jest] npm run test command doesn't work with fresh new install


tonimedina-pri:
Hi guys,

I get the following error when running `npx create-nuxt-app appname && cd appname && npm run test`.

```
 FAIL  test/Logo.spec.js
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/username/Desktop/appname/test/Logo.spec.js:1
    import { mount } from '@vue/test-utils';
    ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      at Runtime._execModule (node_modules/jest-runtime/build/index.js:1166:56)

Running coverage on untested files...Failed to collect coverage from /Users/username/Desktop/appname/pages/index.vue
ERROR: /Users/username/Desktop/appname/pages/index.vue: 'import' and 'export' may only appear at the top level (31:0)

  29 | //
  30 | //
> 31 | import Logo from '~/components/Logo.vue';
     | ^
  32 | export default {
  33 |   components: {
  34 |     Logo
STACK: SyntaxError: /Users/username/Desktop/appname/pages/index.vue: 'import' and 'export' may only appear at the top level (31:0)

  29 | //
  30 | //
> 31 | import Logo from '~/components/Logo.vue';
     | ^
  32 | export default {
  33 |   components: {
  34 |     Logo
    at Parser._raise (/Users/username/Desktop/appname/node_modules/@babel/parser/lib/index.js:754:17)
    at Parser.raiseWithData (/Users/username/Desktop/appname/node_modules/@babel/parser/lib/index.js:747:17)
    at Parser.raise (/Users/username/Desktop/appname/node_modules/@babel/parser/lib/index.js:741:17)
    at Parser.parseStatementContent (/Users/username/Desktop/appname/node_modules/@babel/parser/lib/index.js:11284:18)
[...]

PR #649, opened 2020-10-30
Update Jest package.json to include NODE_ENV

lyndseypadget:
Proposed fix for #553

### DIFF
--- a/packages/cna-template/template/frameworks/jest/package.json
+++ b/packages/cna-template/template/frameworks/jest/package.json
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "test": "jest"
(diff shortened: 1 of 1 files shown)


pi0:
It might be another issue since jest does it itself ([see here](https://github.com/facebook/jest/blob/1205299a25f5b928c96d2962b4793e46265ae0e9/packages/jest-cli/bin/jest.js#L13))

lyndseypadget:
@pi0 Can you provide any insight on the CI failures?  Is this really OS dependent?

clarkdo:
@lyndseypadget Can you please help check if this is still happing to you in latest create-nuxt-app ?

PR #653, opened 2020-11-02
feat(pwa): add default language to nuxt.config

danielroe:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html#Accessibility_concerns

Apologies for picking English, but it seems surprisingly difficult to detect the system locale in a cross-platform manner (without adding a dependency).


pi0:
Can we also set [manifest.lang](https://pwa.nuxtjs.org/manifest) and fallback in case of pwa module not selected?

pi0 (approved):

clarkdo (approved):

PR #660, opened 2020-11-13
fix: added dir: 'ltr' to nuxt.config.js

magyarb:
Nuxt omits `text-align: end;` and `text-align: start;` CSS rules if `dir` attribute is not set. Some frameworks, like Vuetify, use `start` and `end` by default, as these are needed for RTL support.

Details in this [issue.](https://github.com/nuxt/nuxt.js/issues/4555)

This PR adds a default `dir: 'ltr'` setting, solving this issue.


clarkdo (approved):

pi0 (changes requested):
I think this way we are hiding the root cause (also making nuxt.config less minimal). Would you please create an updated repro for issue?

PR #664, opened 2020-11-24
feat: Add types for selected features to tsconfig.json

### LINKED ISSUE #464: Typescript: when using jest for tests, @types/jest should be included
`@types/jest` should be included in `tsconfig.json`'s `types` array
and the library should be pulled as a dependency

Reggino:
For typescript projects, it would be nice if the types for the selected options were added to tsconfig.json automagically.

This fixes #464

### DIFF
--- a/packages/create-nuxt-app/lib/saofile.js
+++ b/packages/create-nuxt-app/lib/saofile.js
@@ -125,6 +125,27 @@ module.exports = {
       }
     })
 
+    if (this.answers.language.includes('ts')) {
+      // Add types for any additional selected package
+      const hasJest = this.answers.test.includes('jest')
+      const hasAxios = this.answers.features.includes('axios')
+      actions.push({
+        type: 'modify',
+        files: 'tsconfig.json',
+        handler (data) {
+          const { compilerOptions } = data
+          const { types } = compilerOptions
+          if (hasJest) {
+            types.push('@types/jest')
+          }
+          if (hasAxios) {
+            types.push('@nuxtjs/axios')
+          }
+          return { ...data, compilerOptions: { ...compilerOptions, types } }
+        }
+      })
+    }
+
     // For compiling package.json
     actions.push({
       type: 'add',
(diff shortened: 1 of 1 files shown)


packages/create-nuxt-app/lib/saofile.js
```diff
@@ -125,6 +125,27 @@ module.exports = {
...
+      const hasJest = this.answers.test.includes('jest')
+      const hasAxios = this.answers.features.includes('axios')
+      actions.push({
+        type: 'modify',
+        files: 'tsconfig.json',
+        handler (data) {
+          const { compilerOptions } = data
+          const { types } = compilerOptions
+          if (hasJest) {
+            types.push('@types/jest')
```
clarkdo, line 139: Should we add `@types/jest` and `@nuxtjs/axios` in corresponding package.json  ?

Reggino: They are actually already there as a sub-dependency, but i'll add them to package.json to make it more explicit.

Reggino: I've added the Jest-types to `` package.json```

```@nuxtjs/axios```is already part of ```package.json``` when axios is selected, so no need for that one.

**Review by clarkdo — APPROVED** [review]

File: packages/create-nuxt-app/lib/saofile.js [hunk 2]
```diff
@@ -117,14 +117,39 @@ module.exports = {
...
+    const hasJest = this.answers.test.includes('jest')
+    const hasAxios = this.answers.features.includes('axios')
     actions.push({
       type: 'modify',
       files: 'package.json',
       handler (data) {
-        return { ...data, ...pkg.load(generator) }
+        const result = { ...data, ...pkg.load(generator) }
+        if (hasTypeScript && hasJest) {
+          result.devDependencies['@types/jest'] = '^26.0.15'
```
clarkdo, line 129: We can put the dep in pacakge.json of `jest` package dir and use `packages/cna-template/template/frameworks/jest/package.js` to remove it when there is no typescript selected.

clarkdo (changes requested):

PR #688, opened 2021-01-12
feat: add nuxt content support for typescript

joukokar:
When creating a project with typescript and nuxt content, the types are missing so using `this.$content(...)` throws type errors.

This fixes it by adding `@nuxt/content` types to tsconfig.json as described here https://content.nuxtjs.org/installation/#typescript

### DIFF
--- a/packages/cna-template/template/tsconfig.json
+++ b/packages/cna-template/template/tsconfig.json
@@ -25,7 +25,12 @@
     },
     "types": [
       "@types/node",
+<%_ if (content) { _%>
+      "@nuxt/types",
+      "@nuxt/content"
+<%_ } else { _%>
       "@nuxt/types"
+<%_ } _%>
     ]
   },
   "exclude": [
--- a/packages/cna-template/template/tsconfig.json
+++ b/packages/cna-template/template/tsconfig.json
@@ -25,12 +25,12 @@
     },
     "types": [
       "@types/node",
-<%_ if (content) { _%>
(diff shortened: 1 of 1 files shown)


Atinux (approved):

packages/cna-template/template/tsconfig.json
```diff
@@ -25,7 +25,12 @@
     },
     "types": [
       "@types/node",
+<%_ if (content) { _%>
```
clarkdo, line 28: indentation can be kept because `<%_` will strip all whitespace before it

joukokar: Fixed the identation

packages/cna-template/template/tsconfig.json
```diff
@@ -25,7 +25,12 @@
     },
     "types": [
       "@types/node",
+      <%_ if (content) { _%>
+      "@nuxt/types",
+      "@nuxt/content"
+      <%_ } else { _%>
       "@nuxt/types"
+      <%_ } _%>
```
danielroe, line 33:
```suggestion
      "@nuxt/types",
      <%_ if (content) { _%>
      "@nuxt/content",
      <%_ } _%>
      "@types/node"
```

joukokar: Hmm. I was under impression the order would matter here, but doesn't seem to matter... especially since nuxt would not extend node

danielroe (approved):

PR #708, opened 2021-02-07
feat: add vant ui framework

themilanfan:


themilanfan:
Not sure why it doesn't fetch the plugin .js file and .css file, would anyone know why?

PR #713, opened 2021-02-10
Change Nuxt+TS logo colors

hsduiii:
Would be nice to use the Nuxt-Typescript logo when choosing ts in create-nuxt-app


hsduiii:
imagen

pi0:
Thanks for PR @hsduiii but i think it will make inconsistency for nuxt logo usage since typescript support is an adhoc feature and progressively we are adding it to core (by v2.15 typescript runtime is built-in and by v3 also built time is out of the box)

hsduiii:
@pi0 Okok, ill close this PR, greetings!

PR #715, opened 2021-02-12
feat(template): update to husky v6

clarkdo:


danielroe:
This would be worth not merging before Husky v5 leaves early access and returns to MIT license.

clarkdo:
@danielroe Agree with your license comment, I'll hold the pr.

packages/cna-template/template/nuxt/.husky/commit-msg
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+<%= pmRun %> commitlint --edit $1
```
danielroe, line 4: yes - nice :-)

Issue #719, opened 2021-02-16
remove `nuxt-ts` as no longer required


danielroe:

clarkdo:
`nuxt-ts` is from `@nuxt/typescript-runtime` not previous `nuxt-ts`, so  I think it should be fine ?

danielroe:
With `jiti`, we can define `serverMiddleware` and `nuxt.config` in Typescript, so we don't require `nuxt-ts` installed by default in new TS projects.

clarkdo:
What about internal nuxt module like `~/modules/socketClient` ?

PR #721, opened 2021-02-17
fix(ts): remove @nuxt/typescript-runtime

clarkdo:
Close https://github.com/nuxt/create-nuxt-app/issues/719

`@nuxt/typescript-runtime` can be removed as nuxt has added `jiti` as ts runtime.


danielroe (approved):

kevinmarrec (approved):
Sounds good :D

pi0 (approved):

PR #734, opened 2021-03-02
fix: Oruga packagePatterns in renovate.json

astagi:
This PR fixes Oruga configuration for renovate.json. I tested it running renovate locally on another repository with the right configuration, I don't know if there's a better way to test renovate.json configuration without creating pull requests.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -54,7 +54,7 @@
       "groupName": "webdriverio packages"
     },
     {
-      "packagePatterns": ["@oruga-ui", "oruga"],
+      "packagePatterns": ["@oruga-ui"],
       "groupName": "oruga packages"
     },
     {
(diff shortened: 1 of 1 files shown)


clarkdo (approved):

renovate.json
```diff
@@ -54,7 +54,7 @@
       "groupName": "webdriverio packages"
     },
     {
-      "packagePatterns": ["oruga"],
+      "packagePatterns": ["@oruga-ui", "oruga"],
```
clarkdo, line 57: Can we remove `oruga` as `@oruga-ui` will cover all scoped packages  ?

astagi: @clarkdo sure! done in a new commit

PR #746, opened 2021-03-10
TypeScript option improvements

loonpwn:
With the latest Nuxt providing out-of-the-box support for the TypeScript runtime and making the dependency and buildModule deprecated, it would be nice if this package didn't add them in.

I did try implementing the config to be generated as https://typescript.nuxtjs.org/cookbook/configuration/, but it seems there is an issue.


loonpwn:
> @loonpwn Note that it is the _runtime_ that should work out-of-the-box; the buildModule is still required for typescript support within a webpack context.

Hey @danielroe. Thanks for checking over it. Looks like I may have got confused, think I was using an earlier version of nuxt-create-app then. Was getting the warning to remove it and saw things were working so assumed I was down the right path.

Oh well, was interesting to look at the code for this. Will close this off

PR #751, opened 2021-03-17
feat: update @nuxtjs/tailwindcss to v4

clarkdo:


packages/cna-template/template/frameworks/tailwind/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "devDependencies": {
-    "@nuxtjs/tailwindcss": "^3.4.3"
+    "@nuxt/postcss8": "^1.1.3",
```
Atinux, line 3: It has to be `postcss8` not `@nuxt/postcss8`

clarkdo: I don't find postcss8 as a valid package in npm registry. https://www.npmjs.com/package/postcss8, do you mean postcss@8 ?

pi0: Yes. It is internally used in tailwind module to force resolving to updated deps. Only to prevent hoisting when a plugin implicitly depending, we should add `postcss@8` and `tailwindcss@2` as dependencies to main project

clarkdo: Do you also mean postcss@8 instead of postcss8 ?

clarkdo: I'm a bit confused about `It has to be postcss8 not @nuxt/postcss8`, I think we need all of these:

```js
postcss@8
tailwindcss@2
nuxt@2.15
@nuxt/postcss8@1.1.3
```

pi0: > @nuxt/postcss8@1.1.3
This one is not neeed. tailwindcss module internally [installs it](https://github.com/nuxt-community/tailwindcss-module/blob/e00da222a62548ad25585ab24d793f596f6c478a/src/index.ts#L37)

Atinux (approved):

PR #762, opened 2021-03-25
perf: use `npm ci` instead of `npm i` in GH Action Workflow

vinayakkulkarni:
- [x] Increases the performance of build time in turns savings lots of $$$ & environment!

MicrosoftTeams-image


clarkdo (approved):

clarkdo:
Thanks!

PR #765, opened 2021-03-30
feat: add support for husky v6

vinayakkulkarni:
Husky is finally MIT again! Wo-Hoo 🥳

This PR will enable support for husky v6 out of the box.

@clarkdo: Pl review & _you can successfully close #715 now_ :)


packages/cna-template/template/frameworks/husky/.husky/pre-commit
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run lint:js
+npm run lint:style
```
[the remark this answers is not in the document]
clarkdo: Yes, please check my original pr

packages/create-nuxt-app/lib/prompts.js
```diff
@@ -115,7 +115,8 @@ module.exports = [
     choices: [
       { name: 'jsconfig.json (Recommended for VS Code if you\'re not using typescript)', value: 'jsconfig.json' },
       { name: 'Semantic Pull Requests', value: 'semantic-pull-requests' },
-      { name: 'Dependabot (For auto-updating dependencies, GitHub only)', value: 'dependabot' }
+      { name: 'Dependabot (For auto-updating dependencies, GitHub only)', value: 'dependabot' },
+      { name: 'Git Hooks (via husky)', value: 'husky' }
```
clarkdo, line 119: I’m not sure if this is a good idea as actually we’re using husky only for lint in git but not provide any other special hooks.

clarkdo (commented):
Thanks for the pr, I was going to update my pr as well.

packages/create-nuxt-app/lib/prompts.js
vinayakkulkarni: I totally leave it upto you & the Nuxt team to decide on the approach, I'm just happy I get to contribute :)

clarkdo: I think create-nuxt-app is more a nuxt template generater instead of generic application solution, so it makes sense to use lint-stage and husky to run vue and nuxt lint rules. And having hubsky as separeted may have some confilcts like:
- Select lint-stage/commitlint but not husky
- Select husky but not lint-stage/commitlint which is just adding husky as dep but no usage at all.

clarkdo: @danielroe What do you think ?

vinayakkulkarni: That's a quite simple use-case of handling whether husky should be enabled or not.

1. https://github.com/nuxt/create-nuxt-app/pull/765/files#diff-c6eb019e77b72c8fb4b0f0974588817517c54196314fee9c36e33281783d520bL35
2. https://github.com/nuxt/create-nuxt-app/pull/765/files#diff-c6eb019e77b72c8fb4b0f0974588817517c54196314fee9c36e33281783d520bL37
3. https://github.com/nuxt/create-nuxt-app/pull/765/files#diff-c6eb019e77b72c8fb4b0f0974588817517c54196314fee9c36e33281783d520bL56

We just have to update/add some logic around this & maybe use [when()](https://github.com/nuxt/create-nuxt-app/blob/master/packages/create-nuxt-app/lib/prompts.js#L123) ? :)

clarkdo: Yes, we can fix the conflicts, my point is husky if set up and installed as dev dependency without any usage if user doesn't choose lint, I'm afraid that we're adding more and more nuxt/vue unrelated logic to cna and we'd better avoid it, cna should focus on nuxt/vue.

danielroe: @clarkdo I tend to agree with you. Anyway, it's a separate question from upgrading Husky to v6. Maybe merge #765 and then make this PR about reconsidering role of Husky?

clarkdo: @danielroe We could, but current pr is adding husky as a new selection `Git Hooks (via husky)`

danielroe: Sorry, I meant #715.

clarkdo: Ah, good point.

clarkdo: @vinayakkulkarni Original pr has been merged, can you please update the pr add title ?

vinayakkulkarni:
Will create new for husky v7

PR #766, opened 2021-04-02
fix(husky): stdin is not a tty in git bash

clarkdo:
See https://typicode.github.io/husky/#/?id=yarn-on-windows


scscgit:
Isn't this only relevant for Yarn? The code is generated even if the user has selected Npm.

Also what's the purpose of `isWindows === true`, does it mean that if the user who generated a project used Linux, then nobody who uses Windows will be able to clone and use it? Is it really a good idea to make this implicit?

clarkdo:
Fair point, pr is always welcome or I’ll make a fix within this week.

scscgit:
PR sent. I was considering whether it'd be better to keep an empty common.sh for any future use when npm is selected, but as this is a workaround created by husky it'd be most likely redundant, so I just removed it. I also included one suggestion of an improvement based on https://github.com/zkat/npx/issues/220.

I wasn't really used to the codebase of this generator, so I hope I tested it correctly. By the way, is there any contributor guide to this project? New users currently have to figure out how to run the project by running `cac` as `.js` via `node`, which most likely requires bootstrapping lerna, and also to assume what's supposed to be covered by the snapshot tests, considering that this husky change seems to have passed all the tests which should fail. I most likely plan to make some other PR in a near future, as there's plenty of things that new Nuxt users still must figure out by themselves, so having any officially maintained contributors guide available would be helpful to avoid having to waste a few days researching it from scratch by trial & error :)

clarkdo:
Thanks for the pr.

Regarding contributor guide, we’ll add to this repo shortly.

scscgit:
[Off-topic] Not related to PR, but regarding the contributor guide, I'll just mention that the SAO will most likely need some explicit guidelines and workarounds in addition to their official docs, as it seems it lacks basic things like file copy and doesn't support EJS features. Here is my today's experience: https://github.com/saojs/sao/issues/176

PR #774, opened 2021-04-18
fix(cna-template): use npx to invoke locally installed binaries

jamesgeorge007:
Follow up of #715

### DIFF
--- a/packages/cna-template/template/nuxt/.husky/commit-msg
+++ b/packages/cna-template/template/nuxt/.husky/commit-msg
@@ -2,4 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 . "$(dirname "$0")/common.sh"
 
-<%= pmRun %> commitlint --edit $1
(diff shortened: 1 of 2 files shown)


clarkdo (commented):
Can you provide more details about this? I don’t think every user has npx installed

jamesgeorge007:
Say; the user opts for `npm` as the package manager.

https://github.com/nuxt/create-nuxt-app/blob/335ef58d3fda63a388d9ce09829e2b7c23741803/packages/create-nuxt-app/lib/saofile.js#L28

`npm run` would be substituted here. However, there isn't a corresponding script as part of the template.

https://github.com/nuxt/create-nuxt-app/blob/335ef58d3fda63a388d9ce09829e2b7c23741803/packages/cna-template/template/nuxt/.husky/pre-commit#L5

Since `npm` `v5.2.0` `npx` is pre-bundled with `npm`.

https://github.com/typicode/husky/tree/main/docs#locally-installed-binaries

clarkdo (approved):

PR #776, opened 2021-04-19
fix(template): update vant template demo

gzg1023:
I created a project based on vant and found that there were errors in the contents of demo, such as using url attribute instead of to, and updating vant's official website address instead of iviewui.
thanks!


clarkdo (commented):
Can you please fix the lint warnings?

gzg1023:
> Can you please fix the lint warnings?

yes

PR #778, opened 2021-04-21
fix(cli): rephrase not empty directory error message

jamesgeorge007:
The user is greeted with the following message if he/she attempts to create a project in the current directory that is non-empty:-

```sh
Can't create . because there's already a non-empty directory . existing in path.
```

This PR aims at rephrasing the above message such that the directory name shows up instead.

### DIFF
--- a/packages/create-nuxt-app/lib/cli.js
+++ b/packages/create-nuxt-app/lib/cli.js
@@ -34,12 +34,15 @@ cli
     if (cliOptions.info) {
       return showEnvInfo()
     }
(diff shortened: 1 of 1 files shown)


clarkdo (approved):

jamesgeorge007:
It would be nice to have a `--force` flag to override this behavior https://github.com/nuxt/create-nuxt-app/issues/556#issuecomment-674555335
@clarkdo WDYT?

clarkdo:
@jamesgeorge007 Sorry for the late reply, the `--force` idea looks good to me, but I think we can use another name which is more related to dir overridden.

PR #792, opened 2021-05-26
feat: update design

Atinux:
Should look like this now:

Screenshot 2021-05-26 at 18 53 33


packages/cna-template/template/nuxt/components/Tutorial.vue
```diff
@@ -0,0 +1,46 @@
+<!-- Please remove this file from your project -->
+<template>
+  <div class="relative flex items-top justify-center min-h-screen bg-gray-100 sm:items-center sm:pt-0">
+    <link href="https://unpkg.com/tailwindcss@^2/dist/tailwind.min.css" rel="stylesheet">
```
pi0, line 4:
```suggestion
    <link href="https://cdnjs.cloudflare.com/ajax/libs/tailwindcss/2.1.2/tailwind.min.css" rel="stylesheet">
```

cdnjs is probably more stable for production usage.

pi0, line 4:
```suggestion
    <link href="https://cdn.jsdelivr.net/npm/tailwindcss@2.1.2/dist/tailwind.min.css" rel="stylesheet">
```

jsdelivr should be more stable for production. Also didn't checked but isn't `base.min.css` enough?

pi0:
Shall we keep `icon.png`? (or conditionally for pwa module) since it is used to generate actual pwa icons as well as favicon.

packages/cna-template/template/nuxt/components/Tutorial.vue
Atinux: I see two review regarding cdnjs or jsdelivr, what would be the good one?

I actually used the one recommended in TailwindCSS docs: https://tailwindcss.com/docs/installation#using-tailwind-via-cdn

Since it is for dev only I think it is fine.

pi0: Both are good and use Cloudflare as a provider (jsdelivr backed by +5 more). I sometimes see downtimes with unpkg (especially since currently using `^2` range it is depending on resolver and we cannot leverage full cache. If you prefer unpkg, I would suggest at least using pinned version to avoid downtime chances and also better cache perf.

pi0:
icon is fine. sorry gh diff was showing `icon.png` removed

clarkdo (approved):
vuetify template also has logo component, should we remove as well ? 

https://github.com/nuxt/create-nuxt-app/blob/master/packages/cna-template/template/frameworks/vuetify/components/Logo.vue

clarkdo:
OK， I’ll update and do some tests tmr

packages/cna-template/template/nuxt/components/Tutorial.vue
pi0: /cc @Atinux

clarkdo:
@Atinux Can you send me svg file of the new logo ?

clarkdo:
@Atinux @pi0 Updated, can you review ?

Atinux:
Good for me @clarkdo

PR #802, opened 2021-06-16
feat: Automatically import component and store in jest.

kght6123:
This has been addressed because it takes time to import individually each time.

### DIFF
--- a/packages/cna-template/template/frameworks/jest/jest.setup.js
+++ b/packages/cna-template/template/frameworks/jest/jest.setup.js
@@ -1,0 +1,2 @@
+// Enable when you want to import components automatically.
+import "./.nuxt/components/plugin.js";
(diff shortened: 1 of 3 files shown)


packages/cna-template/template/frameworks/jest/jest.setup.js
```diff
@@ -0,0 +1,2 @@
+// Enable when you want to import components automatically.
+import "./.nuxt/components/plugin.js";
```
danielroe, line 2: I'm not sure this should be enabled by default.

pi0: Indeed when `components: true` is not provided, this file does not exists... Maybe using `import().catch(()=>{})`? But i think other than this, by default we have `components: true` in default CNA nuxt config

PR #803, opened 2021-06-23
chore: replace babel-eslint with @babel/eslint-parser

yoyo837:


clarkdo (approved):

Ch3rlO:
This update kind of broke things, I wasn't able to generate nuxt project because of an error with something like this 
```Parsing error: No Babel config file detected for (path to the index.vue) Either disable config file checking with requireConfigFile: false, or configure Babel so that it can find the config files
```

clarkdo:
Will look into this in https://github.com/nuxt/create-nuxt-app/issues/805

PR #825, opened 2021-07-15
fix(husky): revert stdin tty workaround of #766 for npm

scscgit:
Git Bash workaround for using tty, which is officially recommended by
husky in their documentation, is only relevant for yarn, not for npm.
Removed isWindows condition make sure projects made by Linux devs are
fully accessible to Windows users. Updated husky in a parent project.
Added --no-install to npx to avoid accidental installation by default
if a project uses different than latest version, but isn't installed.


clarkdo (approved):

clarkdo (approved):

scscgit:
To add a link between PRs for future reference, the `--no-install` change affects #774 where this perspective of using incorrect version wasn't discussed

PR #829, opened 2021-07-18
feat: lintfix and prettier scripts, fix lint-staged coverage

scscgit:
* Add a `lint:prettier` script to provide users with a recommended way to enforce a unified code style in the project, which is less accessible if they have to research the correct usage, or even access it via npx. The intention should be to provide a configuration where linters don't conflict with each other. Also included this in the `lint` script by default to enforce best practices (but anyone is free to remove it). I can't comprehend why new users should be forced to research this on their own; we need a reference solution as a source of truth for example if we want to setup IDE extensions and make sure that they converge on the same code style (definitely not trivial with all the options in VSCode like Vetur, Prettier, Eslint, built-in...).
  * As this can run over all files, we need to specify a `.prettierignore`, especially to exclude directories like `.nuxt`. Due to https://github.com/prettier/prettier/issues/8506, we have to duplicate the `.gitignore` content without being able to include it e.g. by referencing two ignore files. Users may require this file to differ from their `.gitignore`, e.g. if they need to include results of code generators. Note that as there is no analogy of `eslint-disable`, so this file cannot be avoided. As for code re-use (DRY) in the template file, I was unable to import the content via EJS, and the SAO generator lacks a copy action: https://github.com/saojs/sao/issues/176
* Add a `lintfix` script to provide a quick way of fixing any issues in the entire project, especially before they make a commit. Prettier includes `--list-different` to display only changed files (unified with other linters) instead of every single file in the project.
* `create-nuxt-app`: run `lintfix` at the end instead of hard-coding all of the scripts individually; just like what the user is supposed to do on a regular basis.
* Extend `lint-staged` in `package.json` to include prettier by default, where it requires `--ignore-unknown` https://github.com/prettier/prettier/issues/8136, as you can reproduce the issue by trying to commit an unstyled file that's referenced in the `.prettierignore`. Additionally use `--cache` for `eslint` as a recommended performance improvement, which is already assumed as our current `.gitignore` includes the `.eslintcache`.
* Reordered scripts to have linters near each others, and to keep the `prepare` at the end. (Or could we move it to the top?)
* Add `.ts` to supported `eslint` types. I'd personally prefer keeping it there even if the project is JS-only, but for the sake of consistency I only added it for TS projects.
* Add `.scss`, `.sass`, and `.html` to supported `stylelint` types. Fixing CSS `<style>` tags inside HTML definitely sounds like an expected default behavior to me. This list may not be exhaustive, so please make sure it's kept up-to-date with any popular use-cases.
* Modify `lintStaged` conditions to include it even without `eslint`. Previously this meant that together with husky they weren't configured if the user has only chosen `commitlint`, which I believe is a bug. Now it installs even for prettier, but I also allowed it to be installed as an empty block if chosen alone, because the effects of that selection don't make any sense otherwise; who are we to judge if someone wants to lint something else on staging?


packages/cna-template/template/nuxt/.husky/.gitignore
```diff
@@ -1 +1 @@
-_
\ No newline at end of file
+_
```
scscgit, line 1: Part of `lintfix` results thanks to prettier; I think we should integrate it to the template.

packages/cna-template/template/nuxt/package.js
```diff
@@ -12,17 +12,25 @@ module.exports = {
...
     const prettier = linter.includes('prettier')
     const commitlint = linter.includes('commitlint')
     const lintScripts = {
       eslint: '<%= pmRun %> lint:js',
-      stylelint: '<%= pmRun %> lint:style'
+      stylelint: '<%= pmRun %> lint:style',
+      prettier: '<%= pmRun %> lint:prettier'
+    }
+    const lintfixScripts = {
+      eslint: "<%= pmRun %> lint:js <%= pm === 'npm' ? '-- ' : '' %>--fix",
```
scscgit, line 25: I used EJS templates because there already was `<%= pmRun %>`, but I don't see a reason not to `` `use ${variables} instead` ``; especially as any incorrect templates in this JS cause a compilation error, e.g. if you escape `\'` within `"` instead of replacing `'` by `"` and using `'` inside. Also the whitespace-trim didn't seem to be very useful here.

packages/cna-template/template/nuxt/package.js
```diff
@@ -53,15 +66,25 @@ module.exports = {
...
+    const lintfixScript = Object.values(lintfixScripts).join(' && ')
+    if (lintfixScript) {
+      pkg.scripts.lintfix = lintfixScript
+    }
+
+    if (!lintStaged && !commitlint) {
+      delete pkg.devDependencies.husky
+      delete pkg.scripts.prepare
+    } else {
+      // Move prepare to make it the last script
```
scscgit, line 83: Instead of moving, we could also create it here.

packages/create-nuxt-app/test/snapshots/index.test.js.md
```diff
@@ -2365,12 +2374,17 @@ Generated by [AVA](https://avajs.dev).
         'core-js': '^3.15.1',
         nuxt: '^2.15.7',
       },
-      devDependencies: {},
```
scscgit, line 2368: Here the lint-staged was previously installed alone.

scscgit:
I'd also suggest removing the [.prettierrc](https://github.com/nuxt/create-nuxt-app/blob/af164a7640bb81832a26b7176572d70cb3fcae39/packages/cna-template/template/_.prettierrc) and instead moving its content to `package.json` ([as supported by cosmiconfig](https://prettier.io/docs/en/configuration.html)) in order to reduce the unnecessary amount of files in the root directory. Most users won't ever touch the defaults, and there doesn't seem to be any advantage in a separate file anyway (e.g. having JS available):
```
"prettier": {
    "semi": false,
    "singleQuote": true
  },
```

scscgit:
Reordered `prettier` to the beginning of `lintfix` to fix #827.

clarkdo (approved):

PR #839, opened 2021-08-03
feat: added eslintingore template and PWA options

Oluwatemilorun:
When the `PWA` option is selected and also `eslint` lint option, eslint logs an error from the auto generated service worker file `sw.js` due to possible conflicting rules.

## .eslintignore
There is no `.eslintignore` file being generated, so a template was created and which includes auto generated service worker files when PWA option is selected.


clarkdo (approved):

packages/create-nuxt-app/lib/saofile.js
```diff
@@ -103,6 +103,7 @@ module.exports = {
       files: '*',
       filters: {
         '_.eslintrc.js': 'linter.includes("eslint")',
+        '_.eslintignore': 'linter.includes("eslint")',
```
vinayakkulkarni, line 106: Instead of adding another file, can't you add sw.js in gitingore?

Cause there are cases where we use prettier, style lint alongside eslint. This approach would also mean that you have to add `.prettierignore` & `.stylelintignore`

vinayakkulkarni (changes requested):

packages/create-nuxt-app/lib/saofile.js
vinayakkulkarni: Eg: 

Updated ignore - 
https://github.com/vinayakkulkarni/map-my-google-photos/blob/main/.gitignore#L83-L84

Lint scripts - https://github.com/vinayakkulkarni/map-my-google-photos/blob/main/package.json#L11-L18

Oluwatemilorun: When the files are generated, `sw.*` is included in the `.gitignore` already. Although, eslint still runs through and lints the auto generated service worker codes causing the linting error when developing or building.

Oluwatemilorun: The error. `@typescript-eslint` is whats emitting the error.

Screen Shot 2021-08-25 at 3 26 44 AM

This disappears when `sw.*` is included in the `.eslintignore`

PR #853, opened 2021-09-21
feat(cna,template): add src directory prompt

### LINKED ISSUE #243: Include project in src/
Change or add settings to expand the generated project under the src directory.

## motivation

This change not only makes the working directory clearer, but also helps to set up the linters and CIs.

As the project gets larger, changing srcDir can be difficult, so it is worthwhile to add settings to this project.

NozomuIkuta:
Resolve #243 (maybe)

I don't know who mainly manages this repository.
So let me request a review from @Atinux and please change reviewers as needed. 🙏 

@ushironoko

Is this what you meant in the original issue?
If so, I appreciate your review on this PR too.
If not, I appreciate additional explanation in the original issue so that I could help.

---

**Explanation of work**

Most are explained in the code.

Just in case, I checked files in `cna-template/template/frameworks` directory, and I believe there is no relative paths (e.g. `../`) used to refer to other files. This means changing `srcDir` doesn't break any such statements.
If I should add unit tests for combination of `srcDir` and frameworks, I will so let me know.

---

👷‍♂️  Since this is my very first PR for create-nuxt-app tool, I may be making mistakes or missing something. 👷‍♀️


packages/create-nuxt-app/lib/prompts.js
```diff
@@ -107,6 +107,11 @@ module.exports = [
     ],
     default: 'server'
   },
+  {
+    name: 'srcDir',
+    message: 'Src directory:',
```
NozomuIkuta, line 112: Any suggestion for better question message?

clarkdo (approved):
Thanks for the pr, the change looks good, but I may ask a review from @nuxt/framework to see if we should keep a best practice and convention on structure or suggest users defining what ever they want in nuxt projects.

packages/create-nuxt-app/lib/prompts.js
danielroe: What about something like:
```suggestion
    message: 'Source directory, relative to root:',
```

NozomuIkuta: Thank you for suggestion!
I applied your code and modified same kind of wording as well: b1be40b4988e3bef23b56d8c0f8a73a67d7e214f

ushironoko (approved):
This PR seems to be an exact and better implementation of my intended suggestion. I appreciate it.

NozomuIkuta:
Let me close this PR since it's stale.

PR #922, opened 2022-01-05
feat: ignore eslint conflict with keyword files

TerabyteTiger:
## Types of changes

- [ ] Bug fix (a non-breaking change which fixes an issue)
- [X] New feature (a non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Description

By default create-nuxt-app is installing eslint with vue rulesets that throw an error on `.vue` files that are only a single word. This includes when Nuxt layouts are created as outlined [in the Nuxt documentation](https://nuxtjs.org/docs/concepts/views) for custom error pages, and `default.vue` (Default doesn't throw a console error and prevent the app from loading, but it does show an error in VS Code without these changes). 

My change adds a rule to the default eslint files that will prevent these errors from appearing for `default.vue` and `error.vue` files. An addition may want to be added to cover `blog.vue` since that's an example also called out in the docs. 

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes (if not applicable, please state why)

I manually tested to make sure the new eslint files generated would correctly ignore the `default.vue` and `error.vue` files, but I don't believe the tests around the cli are otherwise affected since it's just an extension of the currently generated file (If this is incorrect I'd love to update the associated tests as well 😄 )


clarkdo (commented):
This rule has been disabled in eslint config by https://github.com/nuxt/eslint-config/pull/190, it will be released soon.

TerabyteTiger:
Oh! Good catch 😅

